PR #2, opened 2015-10-07
assessment.js

wiltrahan:
As you can see by this assessment, I'm still lost.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -24,7 +24,13 @@ var sum = function sum() {
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
+  var result = true;
+    array.forEach(function(currentValue, index, array) {
 
+  result = result && callback(currentValue, index, array);
+  });
+
+  return result;
 };
 
 // Question 4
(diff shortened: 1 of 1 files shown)


lib/assessment.js
```diff
@@ -24,7 +24,13 @@ var sum = function sum() {
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
+  var result = true;
+    array.forEach(function(currentValue, index, array) {
 
+  result = result && callback(currentValue, index, array);
```
MattB-GA-Boston: Close, actually! Replace `&&` (AND) with `||` (OR) and your solution will work.

lib/assessment.js
```diff
@@ -32,7 +38,13 @@ var any = function any(array, callback, thisArg) {
 // made up of the array elements for which invoking
 // the callback returns a truthy value
 var select = function select(array, callback, thisArg) {
+  var result = [];
+    array.forEach(function(currentValue, index, array) {
+
+  result = result && callback(currentValue, index, array);
```
MattB-GA-Boston: In this case, you really want to be pushing data into the result array. Does that help?

MattB-GA-Boston:
I agree, it may be worthwhile to spend a little time here. Doing more CodeWars problems may also help you get more comfortable with solving these assessments.

PR #3, opened 2015-10-07
Response to Assessments

baz1389:


lib/assessment.js
```diff
@@ -8,40 +8,48 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
-};
+var min = function min(previousValue, currentValue) {
+  return previousValue < currentValue ? previousValue : currentValue;
+}; //did you mean to use the term reduce here? Reduce takes a set of values - not just for one single (minimum) value.
```
MattB-GA-Boston: That's right! Here's what happens. If we pass this callback to `reduce`, `reduce` will compare the first element to the second; if the first is smaller, the callback will return the first element, so that (on the next iteration) that result can be compared with the second element, etc.

lib/assessment.js
```diff
@@ -8,40 +8,48 @@
...
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(previousValue, currentValue, index, array) {
+  return previousValue + currentValue;
```
MattB-GA-Boston: Yep, that's it exactly.

MattB-GA-Boston:
Nice job on the first two! I recommend seeing if you can work out the remaining three, with a little more time.

PR #4, opened 2015-10-07
Finished assessment.

atsundwn:


MattB-GA-Boston:
You got all of the first four right, so you really should give Q5 a try; let me know if you'd like my feedback on your answer!

PR #5, opened 2015-10-07
"'completed' assessment'"

jcmeyer10:


lib/assessment.js
```diff
@@ -5,18 +5,31 @@
 // its index, and the array itself
 // e.g. callback(currentValue, index, array)
 
+Matt or Saad: this has been a rough day for assessments, I will do whatever it takes to learn this stuff I just need some help.  I feel like I need to start over on the lesson for call backs then move into the four potential array methods.
+
```
MattB-GA-Boston: Do you want to resubmit it? I'd be happy to give you feedback on what you turn in.

Alternatively, we can just work through the solution together. Let me know what works for you!

lib/assessment.js
```diff
@@ -5,18 +5,31 @@
...
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
+var num = [1,3,5,7,9]
 
+var sum = function sum() {
+  num.reduce(sum(previousValue, currentValue, index, array){
+return previousValue + currentValue
+  }
```
MattB-GA-Boston: It's a little convoluted, but your callback is actually right! 
```javascript
var sumCallback = function (prev, curr, index, array) {
  return prev + curr;
};
```

MattB-GA-Boston:
Thanks for submitting anyway, even though you weren't feeling good about your answers. By seeing what things trip you up, we can figure out how to target those areas and get you over your hurdles.

jcmeyer10:
Thanks for the thoughtful response.  I would like to sit down for a few to
talk it through if you have a few minutes.  I can do it during lunch or
after class. Whatever works best for you.

I would also like to chat about this project.  I tried to get some of the
JS done this weekend and am struggling to just get started and make headway
and it's been wreaking havoc on my nerves.  I have never been in a position
like this.  It's one thing to put together a paper in a language I feel
comfortable with.  For me personally it's another to deal with a language I
don't feel comfortable with if that makes sense.

I look at sample code and tend to understand what's going on but putting it
together is another thing.

Thanks again for everything.  I know I am not the first person to go
through this but all I can think right now is the clock ticking down to
Friday and I will be empty handed and then I'll get kicked out and so on
and so forth, you can see where it starts to get mentally heavy.

On Mon, Oct 12, 2015 at 9:26 AM, Matt Brendzel <notifications@github.com>
wrote:

> Thanks for submitting anyway, even though you weren't feeling good about
> your answers. By seeing what things trip you up, we can figure out how to
> target those areas and get you over your hurdles.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ga-wdi-boston/js-array-methods-assessment/pull/5#issuecomment-147399318>
> .
>



-- 
Regards,
Jeff Meyer

857-206-3572

PR #6, opened 2015-10-07
Response

kuramameng:


lib/assessment.js
```diff
@@ -8,15 +8,17 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(p, c, i, a) {
+  var result = p;
+  if(result > c) result = c;
+  return result;
```
MattB-GA-Boston: Yes, this will work (though I'm not sure if you might need curly braces after the `if`). 

Here's another way that it could have been written.
```javascript
var min = function min (prev, curr, index, arr) {
  return (prev < curr)? prev : curr;
};
```

MattB-GA-Boston: Apparently, the curly braces are not _strictly_ necessary; however, I would still defer towards your including them.

lib/assessment.js
```diff
@@ -32,14 +38,24 @@ var any = function any(array, callback, thisArg) {
...
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
 var find = function find(array, callback, thisArg) {
-
+  var result = [];
+  array.forEach(function(c, i, a){
+    if(callback(c, i, a))
+      result.push(c);
+  })
+  return result[0];
```
MattB-GA-Boston: Sure, this will work if there's at least one match. However, if there are no matches, it may act a little funky.The actual `find` will return `-1` if something can't be found. Can you see if there might be a way to do that?

MattB-GA-Boston:
5 of 5

Nice work - keep it up!

PR #8, opened 2015-10-07
responses for array methods assessment

christinemoses:


lib/assessment.js
```diff
@@ -8,15 +8,39 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
+var min = function min(array, callback, initialValue) {
```
MattB-GA-Boston: The question is asking us to create a callback function `min` to pass into reduce as an argument; i.e.
`[1,2,3].reduce(min)`

Here's an example of one that might work.
```javascript
  if (previous < current) {
    return previous;
  }  else {
    return current;
  }
};
```

lib/assessment.js
```diff
@@ -8,15 +8,39 @@
...
 // the sum of the values in the array
 var sum = function sum() {
 
+
+
+// from MDN:
+// var total = [0, 1, 2, 3].reduce(function(a, b) {
+//   return a + b;
+// });
+// // total == 6
```
MattB-GA-Boston: That's exactly right, actually - the callback in there will cause all of the values to get added up.

MattB-GA-Boston:
Thanks for submitting anyway, even if you weren't sure about your answers - it helps us to diagnose and address issues that you might be having. Let me know if you want me to walk through this with you; alternatively, you could resubmit this and I could give you more feedback.

PR #9, opened 2015-10-07
Final commit

gracenjambi:
Used class notes and codes provided in the code along when answering the 3 questions I was able to answer.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -8,14 +8,31 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
+var min = function min(12,9,7) {
+for (var i = 0; i < min.length; i++) {
+    //if the current element is greater than the minValue
+    if (min[i]>minValue) {
+      //replace minValue within the current number
+        minValue = min[i];
+    }
+  };
+  //return the result
+  return minValue;
 
-};
+}
+
+  var result = min([12,9,7]);
+
+  console.log (result);
+
 
 // Question 2
-// Write a callback for reduce that produces
+// Write a  callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
+var sum = function sum([3,5,7,9]) {
+  var sum = [3,5,7.9].reduce(function sum) {
+    return sum;
+  }
 
 };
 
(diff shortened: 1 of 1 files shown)


lib/assessment.js
```diff
@@ -8,14 +8,31 @@
...
+    }
+  };
+  //return the result
+  return minValue;
+
+}
+
+  var result = min([12,9,7]);
+
+  console.log (result);
```
MattB-GA-Boston: The question asks for a callback function, min, which we can pass into reduce, i.e. [1,2,3].reduce(min);

Here's an example of one we might have used.
```javascript
  if (previous < current) {
    return previous;
  } else {
    return current;
  }
}
```

lib/assessment.js
```diff
@@ -8,14 +8,31 @@
...
 
 // Question 2
-// Write a callback for reduce that produces
+// Write a  callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
+var sum = function sum([3,5,7,9]) {
+  var sum = [3,5,7.9].reduce(function sum) {
+    return sum;
+  }
```
MattB-GA-Boston: See my comment above.

lib/assessment.js
```diff
@@ -24,9 +41,16 @@ var sum = function sum() {
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
+  var.result = true;
+  array.forEach(function(currentValue, index, array) {
+
+    result = result && callback(currentValue, index, array);
```
MattB-GA-Boston: Very close! One issue. If we use `&&`, our final result will only be true if _all_ elements satisfy the condition.

Let's work through an example. Say that a hypothetical callback function `isEven` returns `true` for even numbers and `false` for odd numbers. 
Starting at the left of the array `[0,1,2,4,6]`, we would first use the callback to compare `true` and `isEven(0)`; `isEven(0)` is true, so the result (`true && true`) gives the result `true`. We then take that first result, `true`, and compare it with `isEven(1)`. However, `isEven(1)` is _false_, so performing  `true && false` gives a **false** result. This `false` then gets compared with `isEven(2)`, and although `isEven(2)` is `true`, `false && true` still returns `false`. 

In other words, what you're doing is
```javascript
((((true) && isEven(0)) && isEven(1)) && isEven(2)) ....
```
which boils down to
```javascript
true && isEven(0) && isEven(1) && isEven(2) ...
```
This expression will only be true if **all** are true.

However, what we want is to see if **any** are true. To do that, you want to use an 'or' (`||`). 
```javascript
array.forEach(function(previous, current){
  return previous || callback(current);
});
```

MattB-GA-Boston:
1 of 5
Might be worthwhile to try this one again sometime this week. Let me know if you want to submit another pull request - I'd be happy to review it!

PR #10, opened 2015-10-07
completed assessment

raq929:


lib/assessment.js
```diff
@@ -8,15 +8,24 @@
...
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(previousValue, currentValue, index, array) { previousValue = array.slice();
+     array.forEach(function(currentValue, index, array){
+      if (previousValue > currentValue){
+        previousValue = currentValue;
+      }
+     })
+     return previousValue;
```
MattB-GA-Boston: The question asks for a callback function, min, which we can pass into reduce, i.e. [1,2,3].reduce(min);

Here's an example of one we might have used.
```javascript
  if (previous < current) {
    return previous;
  } else {
    return current;
  }
}
```

lib/assessment.js
```diff
@@ -8,15 +8,24 @@
...
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
+var sum = function sum(previousValue, currentValue, index, array) { var sum = 0;
 
+  array.forEach(function(currentValue, index, array){
+    sum += currentValue;
+}
+return sum;
```
MattB-GA-Boston: See comment above. One callback we could have used with `reduce` might be
```javascript
var sum = function(prev, curr) {
  return prev + curr;
};
```

lib/assessment.js
```diff
@@ -24,7 +33,10 @@ var sum = function sum() {
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
-
+ var result = false;
+ array.forEach(function(currentValue, index, array){
+  result = result || callback(currentValue, index, array);
+ }
```
MattB-GA-Boston: Spot on!

lib/assessment.js
```diff
@@ -32,14 +44,26 @@ var any = function any(array, callback, thisArg) {
...
 // the callback returns a truthy value
 var select = function select(array, callback, thisArg) {
-
+  var newArray = [];
+  array.forEach(function(currentValue,index, array){
+    if(callback(currentValue,index, array)){
+      newArray.push(array[index]);
+    }
+  })
+  return newArray;
```
MattB-GA-Boston: Looks good.

lib/assessment.js
```diff
@@ -32,14 +44,26 @@ var any = function any(array, callback, thisArg) {
...
 // for which invoking the callback returns a truthy value
 var find = function find(array, callback, thisArg) {
-
+  var result;
+  array.forEach(function(currentValue,index, array){
+    if (callback(currentValue,index, array)){
+      return result = array[index];
+    }
+  }
+  return result;
```
MattB-GA-Boston: Maybe `return -1` for the last line, so that it gives back -1 if no match is found?

That's being nit-picky, though. You got this.

MattB-GA-Boston:
3 of 5 (-1 on Q1, Q2)

Nice work - let me know if you have any questions about my feedback.

PR #11, opened 2015-10-07
Completed assessment on array objects

sdavidson140:


lib/assessment.js
```diff
@@ -8,23 +8,27 @@
...
 };
 
+
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(previousValue, currentValue) {
+  return previousValue += currentValue;
```
MattB-GA-Boston: Technically will work, but `return previousValue + currentValue` would be more clear.

lib/assessment.js
```diff
@@ -8,23 +8,27 @@
...
+
 // Question 3
 // Write a function that returns true
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
-
+  if (array.every(callback) === true){
+    return true;
+  }
```
MattB-GA-Boston: Hah! Clever, but no.

Think back to the examples above. Is there a way to test if any of the elements pack a particular test? What if we check each of their results?

MattB-GA-Boston:
Whoops... Looks like I never finished your critique. Sorry about that! I'll review the remaining questions now.

lib/assessment.js
```diff
@@ -32,14 +36,20 @@ var any = function any(array, callback, thisArg) {
 // made up of the array elements for which invoking
 // the callback returns a truthy value
 var select = function select(array, callback, thisArg) {
-
+  newArray = [];
+  if (array.forEach(callback) === true) {
+    newArray = array.map(callback);
+  }
+  return newArray;
```
MattB-GA-Boston: You've got a lot of the main elements, but you haven't yet plugged them together. 

Conceptually, what you want to do is
a) create a new array
b) iterate through every element in the old array and
   i) if the element causes the callback to return true, add it to the new array
c) once finished with the iteration, return the new array.

Think you can implement this one now?

lib/assessment.js
```diff
@@ -32,14 +36,20 @@ var any = function any(array, callback, thisArg) {
...
 };
 
 // Question 5
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
 var find = function find(array, callback, thisArg) {
-
+  if (!array.every(callback) === false) {
+    return array[i];
+  }
```
MattB-GA-Boston: Nope, sorry. As above, let's consider a high-level solution in english.

In order to return the first element which matches our condition (i.e. causes our callback to return `true` when passed in as an argument), we must iterate through every element in the array, checking each element against our callback; if we find a match, we end early and return the matching value.

Does that give you any clues as to how to solve this one?

MattB-GA-Boston:
You got the first two - see if you can figure out the other three!

PR #14, opened 2015-10-07
assessment

gmankd:


MattB-GA-Boston:
Pablo, sorry for the holdup in getting you feedback - it looks like there was a miscommunication between me and Saad about who was giving feedback to who. Comments incoming!

lib/assessment.js
```diff
@@ -8,15 +8,19 @@
...
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(array, callback) {
+  for (var i = 0, i <numbers.length; i++) {
+    if(numbers [i]  < min) {
+      min = numbers[i];
+    }
+  }
```
MattB-GA-Boston: This code will give you the minimum value in an array. However, the question was specifically asking you to write a **callback** that could be passed into `.reduce`. 
e.g.
```javascript
var min = function(p, c) {
 ...
};
[1,2,3].reduce(min);
```

lib/assessment.js
```diff
@@ -8,15 +8,19 @@
...
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = [[0,1,2,3].reduce(function sum(a,b) {
+  return a + b});
```
MattB-GA-Boston: See above. That callback is right, though!

lib/assessment.js
```diff
@@ -24,7 +28,11 @@ var sum = function sum() {
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
-
+  var result = true;
+    array.forEach(function(currentValue index, array) {
+      result = result && callback(currentValue, index, array);
+      return result;
+    }
```
MattB-GA-Boston: This could work if `return result` were below the `forEach`, and if you were passing in `array[index]` as an argument to the callback.

lib/assessment.js
```diff
@@ -32,7 +40,24 @@ var any = function any(array, callback, thisArg) {
...
+
+    newArray.push(callback(currentValue, index, array));
+  };
+
+  array.forEach(innerCallback);
+
 
+
+  return newArray;
+};
```
MattB-GA-Boston: Close but not quite. We only want to push a given value into the new array if `callback(value)` is true. Can you see how your code might be modified to get that result?

MattB-GA-Boston:
2 of 5 (50% on all Qs)

You hit close to the mark on all the questions you answered, but weren't quite there - I think you've got the concepts, and just need to work on implementation. If you have any time this weekend, it might be a good idea to see if you could answer the questions now.

Let me know if you want any additional feedback!

PR #16, opened 2015-10-07
assessment

leahphuong:


MattB-GA-Boston:
Leah, sorry for the holdup in getting you feedback - it looks like there was a miscommunication between me and Saad about who was giving feedback to who. Comments incoming!

lib/assessment.js
```diff
@@ -8,24 +8,49 @@
...
-var min = function min() {
+var min = function min(previousValue, currentValue, index, array) {
+
+  var minValue = previousValue;
+
+    if(currentValue < minValue) {
+      minValue = currentValue;
+    }
+  return minValue;
+  }
```
MattB-GA-Boston: Very nice! 

If you wanted to be very terse, you could also write it as follows:
```javascript
var min = function(a, b) {
  return (a < b)? a : b;
};
```

lib/assessment.js
```diff
@@ -8,24 +8,49 @@
...
-};
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
+var sum = function sum(previousValue, currentValue, index, array) {
 
+  var sumTotal = previousValue;
+  return sumTotal += currentValue;
```
MattB-GA-Boston: Yep, this works. Although... it's a bit roundabout.

Why not just write this?
```javascript
var sum = function(a, b){
  return a + b;
};
```

lib/assessment.js
```diff
@@ -8,24 +8,49 @@
...
 // Write a function that returns true
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
-
+  var result = true;
+  array.forEach(function(currentValue,index,array){
+    result = result && callback(currentValue, index, array);
+  });
+  return result;
```
MattB-GA-Boston: Very nice!

MattB-GA-Boston:
3 of 5 - spot on the first three questions. Think you can figure out the last two?

PR #18, opened 2015-10-07
Assessment

kaylal12:
Was not able to figure out question 4 and 5

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -8,15 +8,19 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(previousValue, currentValue) {
+  if (previousValue < currentValue) {
+    return previousValue;
+  } else {
+    return currentValue;
+  }
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(previousValue, currentValue) {
+  return previousValue + currentValue;
 };
 
 // Question 3
@@ -24,7 +28,13 @@ var sum = function sum() {
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
+  var result = true;
 
+  array.forEach(function(currentValue, index, array) {
+    result = callback(currentValue, index, array);
+  })
+
+  return result
 };
 
 // Question 4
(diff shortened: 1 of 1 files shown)


MattB-GA-Boston:
Kayla, sorry for the holdup in getting you feedback - it looks like there was a miscommunication between me and Saad about who was giving feedback to who. Comments incoming!

lib/assessment.js
```diff
@@ -8,15 +8,19 @@
...
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(previousValue, currentValue) {
+  if (previousValue < currentValue) {
+    return previousValue;
+  } else {
+    return currentValue;
+  }
```
MattB-GA-Boston: Yep, this is right.

Here's how it could have been written more tersely.
```javascript
var min = function(a, b) {
  return (a < b)? a : b;
};
```

lib/assessment.js
```diff
@@ -8,15 +8,19 @@
...
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(previousValue, currentValue) {
+  return previousValue + currentValue;
```
MattB-GA-Boston: Correct.

lib/assessment.js
```diff
@@ -24,7 +28,13 @@ var sum = function sum() {
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
+  var result = true;
 
+  array.forEach(function(currentValue, index, array) {
+    result = callback(currentValue, index, array);
+  })
```
MattB-GA-Boston: Not quite. We want this function to give us back a `true` if any element is true; as written, it will only give us the value (true or false) of the last element.

Can you figure out what this would need in order to be correct?
(HINT: Try using `||`).

lib/assessment.js
```diff
@@ -32,13 +42,20 @@ var any = function any(array, callback, thisArg) {
...
+  return newAry;
 };
 
 // Question 5
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
 var find = function find(array, callback, thisArg) {
+  if (callback(currentValue, index, array)) {
+    return currentValue;
+  }
```
MattB-GA-Boston: Actually, this would be correct if you put the code inside a `for` loop. As written, though, it will not work.

lib/assessment.js
```diff
@@ -32,13 +42,20 @@ var any = function any(array, callback, thisArg) {
 // made up of the array elements for which invoking
 // the callback returns a truthy value
 var select = function select(array, callback, thisArg) {
+  var newAry = [];
+
+  newAry = array.forEach(callback(currentValue, index, array))
 
+  return newAry;
```
MattB-GA-Boston: You're right that we need to return a new array. How might we make it so that we only fill that new array with elements for which `callback(element)` gives us `true`?

MattB-GA-Boston:
2.5 of 5 (first two right, plus partial credit on the rest).

It might be a good exercise to see if you can tackle Q3, Q4, and Q5 again now that you've had time to digest the idea of doing these kinds of operations with callbacks. Let me know if you want additional feedback!

PR #19, opened 2015-10-07
Answered all but the last question. ran out of time

gotwilk890:


lib/assessment.js
```diff
@@ -8,14 +8,15 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(a, b) {
+  return a < b ? a : b;
```
srhoulam: Right on.

lib/assessment.js
```diff
@@ -8,14 +8,15 @@
...
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
+var sum = function sum(a, b) {
+  return a + b;
```
srhoulam: That'll do it.

lib/assessment.js
```diff
@@ -24,7 +25,12 @@ var sum = function sum() {
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
+  var result = true;
+  array.forEach(function(currentValue, index, array) {
 
+    result = result && callback(currentValue, index, array);
+  });
+  return result;
```
srhoulam: This is an `every` function, not an `any` function. It tests for whether the callback yields truthy values for *every* element of the list. Here's an explanation and guide to how to turn `every` into `any`:

3/4 credit.

lib/assessment.js
```diff
@@ -32,6 +38,16 @@ var any = function any(array, callback, thisArg) {
 // made up of the array elements for which invoking
 // the callback returns a truthy value
 var select = function select(array, callback, thisArg) {
+  var result = true;
```
srhoulam: You actually don't need a `result` variable in this function. See below.

lib/assessment.js
```diff
@@ -32,6 +38,16 @@ var any = function any(array, callback, thisArg) {
...
+var newArray = [];
+  array.forEach(function(currentValue, index, array) {
+
+    if(result = callback(currentValue, index, array)){
```
srhoulam: This `if` statement's condition is an assignment statement. Fortunately, the variable you're assigning to is actually not important in this function, and the value an assignment statement evaluates to is whatever is on the right side of the assignment operator (`=`). So line 45 here is, aside from the side effect of assigning to `result`, equivalent to `if(callback(currentValue, index, array)) {`. This is good, because it means you're testing for the right thing.

lib/assessment.js
```diff
@@ -32,6 +38,16 @@ var any = function any(array, callback, thisArg) {
...
+      newArray.push(currentValue);
```
srhoulam: Great.

lib/assessment.js
```diff
@@ -32,6 +38,16 @@ var any = function any(array, callback, thisArg) {
...
+    }
+
+  });
+  return newArray;
```
srhoulam: Great, you return the constructed array. This function works correctly. Full credit.

lib/assessment.js
```diff
@@ -39,6 +55,14 @@ var select = function select(array, callback, thisArg) {
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
 var find = function find(array, callback, thisArg) {
+  var result = true;
+  array.forEach(function(currentValue, index, array) {
+
+  if(result = callback(currentValue, index, array)){
```
srhoulam: Same feedback as above applies here. The `result` variable is unnecessary in this function, as well.

lib/assessment.js
```diff
@@ -39,6 +55,14 @@ var select = function select(array, callback, thisArg) {
...
+      return currentValue;
```
srhoulam: Here's some interesting trouble. `find` won't work as intended for a a reason that will help you deepen your understanding of functions.

When you `return` from a function, the flow of control (i.e., the order code gets executed in) returns to the place immediately following where the function was called. In this case, your callback to `Array.forEach` is being called by `Array.forEach`, so it returns to `Array.forEach` with the value. `Array.forEach` doesn't do anything with the value and just keeps iterating through the array. `Array.forEach` also doesn't return a value, so any values to returned by the callback you give it are simply lost.

As a result, `find` always returns `undefined`. For this question, a classic `for` loop containing the `if` statement you constructed would have worked. 3/4 credit.

srhoulam:
4.5/5. Rock on.

PR #20, opened 2015-10-07
added response to challenge.js file

resource11:


lib/assessment.js
```diff
@@ -8,38 +8,57 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(previousValue, currentValue, index, array) {
+  return currentValue > previousValue ? currentValue : previousValue;
```
srhoulam: This is actually `max`, not `min`. It selects for the greater element; used with `Array.reduce`, it will get you the greatest element in the list. Half credit.

lib/assessment.js
```diff
@@ -8,38 +8,57 @@
...
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(previousValue, currentValue, index, array) {
+  return previousValue + currentValue;
```
srhoulam: Yep, that'll do it.

lib/assessment.js
resource11: Ahh. Yes. Hazard of reading the question too quickly. :) At least the return statements uses the ternary operator. So... style points!

lib/assessment.js
```diff
@@ -8,38 +8,57 @@
...
 // in the array returns a truthy value
-var any = function any(array, callback, thisArg) {
 
+// this seems vague, and I don't exactly know
+// what's being asked here, so I'm making a function
+// that uses a for loop to set every item in the array
+// to be 1
+var any = function any(array, callback, thisArg) {
+  for (var i = 0; i < array; i++) {
+    callback(array[i], i, array) = 1;
```
srhoulam: This line is trouble! Line 33 here is trying to assign to the *return value* of `callback(array[i], i, array)`, which we know is some primitive value that can be coerced into a boolean. Assigning to a value is syntactically incorrect!

For the goal you stated, a correct line 33 would read:

`array[i] = 1;`

Also, on line 32, you are testing `i`, a variable containing a number, against `array`, which presumably contains an array. I think what you meant as your loop condition was `i < array.length`.

1/4 credit.

lib/assessment.js
```diff
@@ -8,38 +8,57 @@
...
 // the callback returns a truthy value
-var select = function select(array, callback, thisArg) {
 
+// Stil not clear. So, this is what I've got.
+var select = function select(array, callback, thisArg) {
+  var newArray = [];
+  for (var i = 0; i < array; i++) {
+    newArray[i] = array[i]
+  }
+  return true;
```
srhoulam: This is close. To do it this way, using a classic `for` loop, you'd need to keep multiple loop indices. On the other hand, if you use `Array.forEach` and `Array.push`, you wouldn't need any index variables.

The problem I pointed out earlier regarding comparing a number to an array also applies here. If this loop were well-formed, it would copy every reference in the source array to the newly-created target array, then lose the reference to the target array and return a boolean instead.

Nonetheless, this is a good start. If you apply one of the approaches mentioned in the first paragraph and use an `if` statement to test for a truthy return value from invoking the callback on the current list element, index, and array, this would be a good answer. 1/2 credit.

lib/assessment.js
```diff
@@ -8,38 +8,57 @@
...
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
-var find = function find(array, callback, thisArg) {
 
+// again, not clear what's wanted here but it looks like
+// using the properties of the .forEach method
+var find = function find(array, callback, thisArg) {
+  var result = 0;
+  if (array[i]!= null) {
+    return array[i];
```
srhoulam: `i` is undefined in `find`'s scope, meaning it'll look out to its parent and ancestor scopes to find (the wrong) `i`. It implies you intended to loop, though. If this `if` statement and its body were in a loop, this function would search for the first non-null value in the array, and return it. It's not a far step from there to then test for the return value of invoking the callback on its expected arguments instead of the array at index `i` being non-null, making this function correct.

Finally, you've got unmatched curly braces. I see two opening curly braces, one for the function body and one for the `if` statement body, and only one closing curly brace below. 1/4 credit.

srhoulam:
2.5/5

PR #21, opened 2015-10-07
assessment response

faetea:
referenced my notes from class
didn't finish

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -4,27 +4,61 @@
 // be called with the current element in the array,
 // its index, and the array itself
 // e.g. callback(currentValue, index, array)
+var reduce = function reduce(array, callback, initialValue) {
+  var previousValue;
+  if (initialValue !== undefined) {
+    previousValue = initialValue;
+  } else {
+    array = array.slice();
+    previousValue = array.shift();
+  }
+
+  array.forEach(function(currentValue, index, array) {
+    previousValue = callback(previousValue, currentValue, index, array);
+  });
+
+  return previousValue;
+};
+
+var numbers = [3, 42, 5, 74, 11];
 
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(previousValue, currentValue, index, array) {
+  return previousValue < currentValue ? previousValue : currentValue;
 };
+
+console.log(reduce(numbers, min));
+
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(previousValue, currentValue, index, array) {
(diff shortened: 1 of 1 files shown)


lib/assessment.js
```diff
@@ -4,27 +4,61 @@
...
+    array = array.slice();
+    previousValue = array.shift();
+  }
+
+  array.forEach(function(currentValue, index, array) {
+    previousValue = callback(previousValue, currentValue, index, array);
+  });
+
+  return previousValue;
+};
```
srhoulam: This is the function that takes the callback. The question wanted you to write the callback that can be passed to `reduce` to make it behave as `Math.min` on an array.

lib/assessment.js
```diff
@@ -4,27 +4,61 @@
...
+
+var numbers = [3, 42, 5, 74, 11];
 
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(previousValue, currentValue, index, array) {
+  return previousValue < currentValue ? previousValue : currentValue;
```
srhoulam: Great. That's what was asked for.

lib/assessment.js
```diff
@@ -4,27 +4,61 @@
...
+console.log(reduce(numbers, min));
+
+
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(previousValue, currentValue, index, array) {
+  return previousValue + currentValue;
```
srhoulam: Cool. This is what was asked for.

lib/assessment.js
```diff
@@ -4,27 +4,61 @@
...
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
+  var result = true;
+  array.forEach(function(currentValue, index, array) {
+    result = result && callback(currentValue, index, array);
+  });
 
+  return result;
+};
```
srhoulam: This is actually `every`, not `any`. It returns `true` only if **all** items in the list (array) yield `true` when passed to the callback. Here is an explanation of the steps required to turn `every` into `any`:

3/4 credit.

lib/assessment.js
```diff
@@ -39,9 +73,38 @@ var select = function select(array, callback, thisArg) {
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
 var find = function find(array, callback, thisArg) {
+  var result = true;
+  array.forEach(function(currentValue, index, array) {
+    result = result && callback(currentValue, index, array);
+  });
 
+  return result;
```
srhoulam: This is `every` again, and it's not necessarily helpful for completing this question. It's good that you made an effort, though. No credit.

srhoulam:
2.75/5

PR #22, opened 2015-10-07
Assessment in progress

kelvinma:
It has become painfully clear that I do not have a firm grasp on how to invoke a callback function... 

I was under the impression that invoking the Array Methods of .filter(), .reduce(), etc., was invoking the callback functions of the methods themselves.

*Update: Upon review of the solution, I invoked the callback as an anonymous function within the starter code, which is why the code worked when tested, just not in the specified manner. The specifics of what the questions were asking for were not clear to me.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -8,15 +8,21 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(array) {
+  var minimum = array.reduce(function(a,b){
+    return b < a ? b : a;
+  });
+  return minimum;
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(array) {
+  var total = array.reduce(function(a,b) {
+    return a + b;
+  });
+  return total;
 };
 
 // Question 3
@@ -24,7 +30,11 @@ var sum = function sum() {
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
-
+  result = true;
+  array.forEach(function(a){
+    result = result && callback(a);
+  });
+  return result;
 };
 
 // Question 4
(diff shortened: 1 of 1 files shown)


lib/assessment.js
```diff
@@ -8,15 +8,21 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(array) {
+  var minimum = array.reduce(function(a,b){
+    return b < a ? b : a;
+  });
+  return minimum;
```
srhoulam: This is not the correct solution, in that it can't be passed in as a callback to `Array.reduce`, but it does *contain* the correct solution.

Assign the following to `min` and you're golden.
```js
function(a,b){
    return b < a ? b : a;
}
```

Full credit for containing the correct answer as an anonymous function.

lib/assessment.js
```diff
@@ -8,15 +8,21 @@
...
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(array) {
+  var total = array.reduce(function(a,b) {
+    return a + b;
+  });
+  return total;
```
srhoulam: Same as above.

lib/assessment.js
```diff
@@ -24,7 +30,11 @@ var sum = function sum() {
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
-
+  result = true;
+  array.forEach(function(a){
+    result = result && callback(a);
+  });
+  return result;
```
srhoulam: This is an `every` function, not an `any` function. Here is an explanation of `any` versus `every`, and the steps needed to convert `every` to `any`:

3/4 credit.

lib/assessment.js
```diff
@@ -32,14 +42,14 @@ var any = function any(array, callback, thisArg) {
 // made up of the array elements for which invoking
 // the callback returns a truthy value
 var select = function select(array, callback, thisArg) {
-
+  return array.filter(callback);
```
srhoulam: Using the built-in array method in your implementation of the same array method is cheating! It's good that you know about it and how to use it, though. No credit.

lib/assessment.js
```diff
@@ -32,14 +42,14 @@ var any = function any(array, callback, thisArg) {
...
 };
 
 // Question 5
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
 var find = function find(array, callback, thisArg) {
-
+  return array.filter(callback)[0];
```
srhoulam: Same as above, except this shows that you know the relationship between `find` and `select`. Doing it this way is wasteful, though, since you build the whole list just to extract its zeroth element. 1/4 credit.

srhoulam:
3/5. Well done, this was a tough assessment.

PR #23, opened 2015-10-07
Array Methods assessment.js updated

bbarbersox:


lib/assessment.js
```diff
@@ -8,33 +8,87 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
+var min = function min() {previousValue, currentValue, index, array)
```
srhoulam: Your `min` function is right-on except for a bit of syntax. This line I have just called out should be:

`var min = function min(previousValue, currentValue, index, array) {`

Note the brackets. Previously, you had an empty set of parens followed by a mismatched curly-paren pair, and no open curly bracket at the end to match the closing curly bracket after the function body.

Here is the template for function syntax:

```js
function name(param1, param2, ..., paramN) {
    statement1;
    statement2;
    ...
    statementN;
}
```

9/10 credit for being correct except for syntax.

lib/assessment.js
```diff
@@ -8,33 +8,87 @@
...
 // Write a function that returns true
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
+  var result = true;
+  array.forEach(function(currentValue, index, array) {
+    // if callback ever returns false, result will be set
+    // to false amd the callback will not be called in any
+    // subsequent iterations.
+    result = result || callback(currentValue, index, array);
```
srhoulam: Great, you noted that the principal difference between `any` and `every` is that any should be ORing the boolean results into an accumulator variable, not ANDing them. However, `result`'s initial value being `true` breaks this, because `true` OR anything yields `true`. So, if you change line 62 so that it initializes `result` to `false`, your `any` function will be entirely correct. 9/10 credit.

lib/assessment.js
```diff
@@ -8,33 +8,87 @@
...
+  return result;
 };
 
+
 // Question 4
 // Write a function that returns a new array
 // made up of the array elements for which invoking
 // the callback returns a truthy value
 var select = function select(array, callback, thisArg) {
+  var newArray = [];
```
srhoulam: Here, we want to use the fact that our callback should return a boolean for each element we pass to it (if it doesn't, that means the user is wrong, not us!) to determine whether we should add that element to the new list we're building.

srhoulam:
2.8/5

PR #24, opened 2015-10-07
Submitted Array Methods Assessment

LaurisaNeuwirth:


lib/assessment.js
```diff
@@ -8,25 +8,80 @@
...
+console.log(numbers.reduce(function (previousValue, currentValue) {
+  return previousValue < currentValue ? previousValue : currentValue;
+}));
+
+
+};
+
+// I used Antony's Max Example of code and switched the operators.
+// I do not think I could logic out the whole thing as concisely as this.
+// I cannot write this code from scratch myself but I can read this and edit it
```
srhoulam: We don't need your code to be as concise as Antony's. He has been doing this for decades!

lib/assessment.js
```diff
@@ -8,25 +8,80 @@
...
+  return previousValue + currentValue;
+}));
+
+
 };
 
+// Same case for solving here as above-
+// I heavily relied on Antony's code and can readi it pretty well
+// I Feel rushed so maybe I am not catching all of the edits but
+// I am concerned that I cannot construct this code myself,
```
srhoulam: You're not expected to be able to write *his* code. You're expected to be able to write less elegant code that still works. This function works, but I'm not clear on what edits you've made to Antony's code.

lib/assessment.js
```diff
@@ -8,25 +8,80 @@
...
+  var result = true;
+  array.forEach(function(currentValue, index, array) {
+    // if callback ever returns false, result will be set
+    // to false amd the callback will not be called in any
+    // subsequent iterations.
+    result = result && callback(currentValue, index, array);
+  });
+
+  return result;
 };
```
srhoulam: This is the body of Antony's `every` function, which is a good starting point. `any` differs from `every` in that we want one `true` to win out over an indeterminate number of `false`s, whereas `every` wants one `false` to win out over an indeterminate number of `true`s. This means using a different starting value for our boolean accumulator variable and a different logical operation.

Why don't we work through this, and more general problem solving approaches, after 4:30 today.

srhoulam: Here is an explanation for how to turn `every` into `any`:

3/4 credit.

srhoulam:
2.75/5

PR #25, opened 2015-10-07
responded to questioins

TeoPaolucci:
Questions were somewhat unclear as to what was required of them.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -8,15 +8,24 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(lastResult, thisNum) {
+  if (!lastResult) {
+    return thisNum;
+  } else if (lastResult < thisNum) {
+    return lastResult;
+  } else {
+    return thisNum;
+  }
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(lastResult, thisNum) {
+  if (!lastResult) {
+    lastResult = 0;
+  }
+  return lastResult + thisNum;
 };
 
 // Question 3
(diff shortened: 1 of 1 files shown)


lib/assessment.js
```diff
@@ -8,15 +8,24 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(lastResult, thisNum) {
+  if (!lastResult) {
```
srhoulam: This is an interesting approach. I don't see why it shouldn't work with a few changes.

Here's an interesting case to consider. Suppose `min` is invoked with arguments `0` and some positive number, e.g., `min(0, 111)`. Zero is a falsey number, so this first `if` condition succeeds and `min` returns the second argument, `111`.

If you're testing for `lastResult` being undefined or null, compare it directly to them in the `if` condition: `if(lastResult === null || lastResult === undefined)`. 3/4 credit.

lib/assessment.js
```diff
@@ -8,15 +8,24 @@
...
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(lastResult, thisNum) {
+  if (!lastResult) {
+    lastResult = 0;
+  }
+  return lastResult + thisNum;
```
srhoulam: Great. This one looks like it'll behave correctly for all numeric inputs.

lib/assessment.js
```diff
@@ -24,7 +33,9 @@ var sum = function sum() {
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
-
+  var ifAny = false;
+  //for each element in the array, use callback, if callback returns true set ifAny to true
+  return ifAny;
```
srhoulam: I'll give this 1/4 credit for correct pseudocode and a good starting value for `ifAny`.

srhoulam:
2.25/5

PR #27, opened 2015-10-07
completed assessment

cwcrane:


lib/assessment.js
```diff
@@ -8,15 +8,21 @@
...
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(previousValue, currentValue, index, array) {
+  var minimum = previousValue;
+  for (var i=0; i<array.length; i++){
+    if (currentValue < minimum){
+      minimum = currentValue;
+    };
+  }return minimum;
```
srhoulam: What keeps your loop from working is that the loop index variable is not used inside the body. If the body of a loop is not dependent on the loop index variable, then it does the exact same thing each time without variation. In this case, it tests the exact same `currentValue` against `minimum` and, if the test succeeds, it assigns the exact same `currentValue` to `minimum`.

Now, that's what is making the loop inoperative. The real problem here is that you shouldn't be looping in the first place. This is meant to be a callback for the `Array.forEach` method, which does all the looping for you. The callback's role is simply to test the parameters for certain conditions (e.g., `currentValue < previousValue`) and return `true` for one case and `false` otherwise.

Half credit.

lib/assessment.js
```diff
@@ -8,15 +8,21 @@
...
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(currentValue, index, array) {
+       currentValue += currentValue;
+     return currentValue;
```
srhoulam: This callback is almost correct, and presents some opportunities for very helpful feedback.

Your `sum` function does not take a `previousValue` parameter. This is very significant because `previousValue` is the *first* parameter. It's not that it's not taking it; it's that it's naming it wrong. Here's an illustration to put the following explanation in context:

```js
var f = function f(a, b, c) {
    return a + b;
}
var g = function g(b, c) {
    return b + c;
}

f(1, 2, 3); // returns 3
g(1, 2, 3) // returns 3
```

If you remove the first parameter from the list, what happens is that the first argument gets assigned to the first parameter still on the list. So the value passed to `sum` that was intended to be assigned to `previousValue` is assigned to `currentValue`, and the value intended to be assigned to `currentValue` is assigned to `index`. Above, this is represented by `g` having a parameter list of two, `b` and `c`. That doesn't mean it ignores the `1` we pass in when we invoke it. It just means it'll refer to `1` as `b` and `2` as `c` when it runs.

Now, in the body of `sum`, what it does is add `currentValue` to itself (effectively multiplying it by two) and assign the result to `currentValue`. It then returns `currentValue`. Since `currentValue` is actually a misnamed parameter that refers to the previously returned value, what this function actually does is repeatedly double the first element (or initial value passed in to `Array.reduce`). Specifically, what it does is multiply the array's zeroth element, or initial value passed in to `Array.reduce`, by 2^array.length

srhoulam:
1/5

PR #28, opened 2015-10-07
literally got no where

francescadal:
ii found this really difficult.


srhoulam:
Why don't we work on this together after 4:30 so you can resubmit this assessment. There's a lot of material here that is very important to your success in this course.

PR #29, opened 2015-10-07
assessment

drewster317:


lib/assessment.js
```diff
@@ -8,32 +8,73 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
+var min = function min(previousValue, currentValue, index, array) {
+    var result = previousValue;
+    if(currentValue < result) result = currentValue;
+    return result;
+
```
srhoulam: Great.

lib/assessment.js
```diff
@@ -8,32 +8,73 @@
...
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
+var sum = function sum(previousValue, currentValue, index, array) {
+       var result = previousValue;
+     result += currentValue;
+    return result;
+
```
srhoulam: Good. The code is correct. All I can ask you to do here is use consistent indentation for future reading.

lib/assessment.js
```diff
@@ -8,32 +8,73 @@
...
+     result = result && callback(currentValue, index, array);
+ });
+ return result;
+
+};
+
+var numbers = [6, 8, 26, 80];
+
+var even = function(currentValue) {
+  return (numbers % 2) === 0;
```
srhoulam: In these callbacks intended for any, you're trying to access the array `numbers` through an outside reference rather than using the passed-in value. Not only do they not work, but if they did, they'd operate on an array other than the one that was passed in to `any` for it to operate on.

lib/assessment.js
```diff
@@ -8,32 +8,73 @@
...
 var any = function any(array, callback, thisArg) {
 
+ var result = true;
+
+ array.forEach(function(currentValue, index, array){
+     result = result && callback(currentValue, index, array);
+ });
+ return result;
+
+};
```
srhoulam: This is a great answer and shows a command of JavaScript. However, `any`'s logic is incorrect.

3/4 credit.

srhoulam:
2.75/5

PR #30, opened 2015-10-07
Answered q's

HoldenMills:


lib/assessment.js
```diff
@@ -8,30 +8,44 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
-};
+var min = reduce(function(previousValue, currentValue, index, array) {
+  return previousValue < currentValue ? previousValue : currentValue;
+)};
```
srhoulam: So, here is the parts of this answer that would be considered good (as in full-points):

```js
var min = function(previousValue, currentValue, index, array) {
};
```

Remember, you were asked to write this function so it can be passed to `Array.reduce` by someone else. That means it's not your responsibility to be invoking the reduce operation in your answer.

There's also a syntax note: sets of brackets can nest, but cannot overlap. That is to say, this is fine: `{ ... ( ... [ ... ] ... ) ... }`, but this is not: `( ... { ... ) ... }`. It's not acceptable for the outer set of brackets to close before all sets of brackets opened within it are closed.

3/4 credit.

lib/assessment.js
```diff
@@ -8,30 +8,44 @@
...
 // in the array returns a truthy value
-var any = function any(array, callback, thisArg) {
+  var every = function every(array, callback, thisArg) {
+  var result = true;
+  array.forEach(function(currentValue, index, array) {
+
+    result = result && callback(currentValue, index, array);
+  });
 
+  return result;
```
srhoulam: This is a good `every` function, but a bad `any` function. What `every` does is start out assuming something is true for all elements in an array, and tries to prove itself wrong. So `result` begins as `true` and the first `false` you AND with it makes it `false`. Nothing ANDed with `false` will ever yield `true`, so it stays `false` indefinitely afterwards.

For `any`, we want to use different logic. We want to start out assuming `false` and OR the return value of the callback with it as we apply it on each array element. Once we OR `true` with it, `result` now stores `true`, and anything ORed with `true` yields `true`. This allows us to measure that at least one element of the array causes the callback to return `true`.

3/4 credit

lib/assessment.js
```diff
@@ -8,30 +8,44 @@
...
 // made up of the array elements for which invoking
 // the callback returns a truthy value
 var select = function select(array, callback, thisArg) {
+  var result = true;
+  array.push(function(currentValue, index, array) {
+
+    result = result && callback(currentValue, index, array);
+  });
+
+  return result;
```
srhoulam: The body of `every` is not a great starting point for this function.

You want to create a new variable inside the function and initialize it with an empty array. Then, as you're iterating over the array, if `callback(currentValue, index, array)` evaluates to `true`, you want to insert it into your newly created array using an array method, such as `push`. Finally, you return the variable your created that points to the array of selected elements you just built.

No credit.

srhoulam:
1.5/5

PR #33, opened 2015-10-13
resubmit

toymechqm:


lib/assessment.js
```diff
@@ -8,28 +8,66 @@
...
+    array = array.slice();
+    previousValue = array.shift();
+  }
+
+  array.forEach(function(currentValue, index, array) {
+    previousValue = callback(previousValue, currentValue, index, array);
+  });
+
+  return previousValue;
+};
```
srhoulam: This is Antony's `reduce` implementation. It's not what the question asked for; Q1 wants the function that will be passed in as `callback` in order to make it return the minimum element of an array of numbers.

lib/assessment.js
```diff
@@ -8,28 +8,66 @@
...
 // Write a callback for reduce that produces
 // the sum of the values in the array
+<<<<<<< HEAD
+var sum = function sum(numbers) {
+    var total = 0;
+for ( var i = 0; i < numbers.length; i++ ){
+    total += numbers[i];
+}
+    return total;
+}
```
srhoulam: `sum` will compute the sum of the array elements effectively when passed the array. It works as a standalone function that takes an array as its only argument.

Q2 asked for a function that gets passed in to `Array.reduce` to operate on at least two arguments, `previousValue` and `currentValue`. It should not loop over anything in its body, as `Array.reduce` is already iterating over the array.

Half credit for writing a good standalone function to do the desired task.

lib/assessment.js
```diff
@@ -8,28 +8,66 @@
...
 
 // Question 3
 // Write a function that returns true
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
+<<<<<<< HEAD
+// No idea how to begin
+// a for loop?
+//I am not understanding callbacks
```
srhoulam: This is a great topic to bring up in a one-on-one or talk to with me or Matt after 4:30 today. Callbacks are tricky and we don't expect anyone to understand them intuitively.

lib/assessment.js
```diff
@@ -37,6 +75,12 @@ var any = function any(array, callback, thisArg) {
 // made up of the array elements for which invoking
 // the callback returns a truthy value
 var select = function select(array, callback, thisArg) {
+<<<<<<< HEAD
+var array = []
+array.map
+// No idea how to begin
+// .map() ????????
```
srhoulam: Here, you are being asked to implement `Array.filter`. It takes a list and a callback, and returns a list with length up to the length of the passed-in list. The callback is used to determine which elements to keep and which to discard.

If I'm not helping someone else after 4:30, I'd be happy to guide you through writing this and clearing up what questions you have. If not, please ask Matt about this during 4:30-5:30, or take advantage of Tom's after-class sessions when he's holding them.

srhoulam:
.5/5

PR #36, opened 2015-10-26
Array

gmankd:
Reanswered

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -8,32 +8,27 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min(array, callback) {
-  for (var i = 0, i <numbers.length; i++) {
-    if(numbers [i]  < min) {
-      min = numbers[i];
-    }
-  }
+var min = function min(previousValue, currentValue) {
+  return previousValue < currentValue ? previousValue : currentValue;
 };
-
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 var sum = [[0,1,2,3].reduce(function sum(a,b) {
-  return a + b});
-};
+return a + b})];
+
 
 // Question 3
 // Write a function that returns true
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
-  var result = true;
-    array.forEach(function(currentValue index, array) {
-      result = result && callback(currentValue, index, array);
+  var result = false;
+    array.forEach(function(currentValue, index, array) {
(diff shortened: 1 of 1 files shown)


lib/assessment.js
```diff
@@ -8,38 +8,64 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(previousValue, currentValue) {
+  return previousValue < currentValue ? previousValue : currentValue;
```
srhoulam: Good. 1

lib/assessment.js
```diff
@@ -8,38 +8,64 @@
...
 };
-
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
+var sum = [[0,1,2,3].reduce(function sum(a,b) {
+return a + b})];
```
srhoulam: The function passed as an argument to `reduce`:

```js
function sum(a,b) {
    return a + b
}
```

is a correct answer to this question. However, the variable `sum` stores a number, which is of the wrong type to be an answer to this question.

1

lib/assessment.js
```diff
@@ -8,38 +8,64 @@
...
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
-
-};
+  var result = false;
+    array.forEach(function(currentValue, index, array) {
+      result = result || callback(currentValue, index, array);
+    });
+      return result;
+    };
```
srhoulam: Yep. 1

lib/assessment.js
```diff
@@ -8,38 +8,64 @@
...
+
+    newArray.push(callback(currentValue, index, array));
+  };
+
+  array.forEach(innerCallback);
+
+
+
+  return newArray;
+};
```
srhoulam: This is not a filter function, but a map function. It will produce a new array containing the booleans returned by `callback` for the elements in the original array.

A filter function uses the `callback`'s return values to determine whether to add the original array element to the new array; it does not store the booleans anywhere.

1/2

lib/assessment.js
```diff
@@ -8,38 +8,64 @@
...
 // for which invoking the callback returns a truthy value
 var find = function find(array, callback, thisArg) {
+var result;
+array.forEach(function(currentValue, index, array) {
+    if (callback(currentValue, index, array)) {
+      result = currentValue;
+    }
+  });
 
+  return result;
```
srhoulam: Great. 1

srhoulam:
4.25/5

PR #39, opened 2015-10-29
Tried this one again...

AkJones2007:


lib/assessment.js
```diff
@@ -8,15 +8,20 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(memo, current) {
+  if (current < memo) {
+    memo = current;
```
srhoulam: Don't assign to parameters! We do this only in exceptional cases. In general, variables are cheap, so use them!

lib/assessment.js
```diff
@@ -8,15 +8,20 @@
...
+  }
+  return memo;
```
srhoulam: Would benefit a lot from not assigning to a parameter, but it is correct. 1

lib/assessment.js
```diff
@@ -8,15 +8,20 @@
...
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(memo, current) {
+  memo = memo || 0;
```
srhoulam: If `memo` (or `previousValue`) is `undefined`, that's not the callback's fault. It's on the developer writing the `reduce` call to set an initial value when needed.

lib/assessment.js
```diff
@@ -8,15 +8,20 @@
...
+  memo += current;
+  return memo;
```
srhoulam: Correct callback, but would be better if it didn't assign to one of its parameters. 1

lib/assessment.js
```diff
@@ -24,7 +29,18 @@ var sum = function sum() {
...
+  array.forEach(function(value) {
+    // Test each value by passing it to the callback
+    if (callback(value)) {
+      // If the callback returns a truthy value,
+      // switch 'isTrue' to the opposite
+      isTrue = true;
+    }
+  })
+  // Return 'isTrue'
+  return isTrue;
```
srhoulam: Well done. 1

lib/assessment.js
```diff
@@ -32,14 +48,27 @@ var any = function any(array, callback, thisArg) {
...
+  // Iterate through each element in an array
+  array.forEach(function(value) {
+    // Test each value by passing it to the predicate
+    if(callback(value)) {
+      // If the predicate returns true, add it to the new array
+      filtered.push(value);
+    }
+  });
+  // Return the filtered array
+  return filtered;
```
srhoulam: Awesome. 1

lib/assessment.js
```diff
@@ -32,14 +48,27 @@ var any = function any(array, callback, thisArg) {
...
 
 // Question 5
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
 var find = function find(array, callback, thisArg) {
-
+  // Collect all truthy values in an array
+  var truthy_values = select(array, callback);
+  // Return the first element of the array
+  return truthy_values[0];
```
srhoulam: Good. You saw that this could be implemented using code you had already written. 1

srhoulam:
5/5

PR #40, opened 2015-10-30
reworked assessment from clean copy

resource11:


lib/assessment.js
```diff
@@ -8,15 +8,45 @@
...
 // the minimum value in the array
-var min = function min() {
 
+// iterate over the array, use ternary operator to compare currentValue
+// to previousValue and assign the min to whichever satisfies
+// the condition
+var min = function min(previousValue, currentValue, index, array) {
+  // set the intial result
+  var result = previousValue;
+  for (var i = 0; i < array.length; i++) {
```
srhoulam: This function is meant to be called by a function that handles iterating over the array for you. That's where you get `previousValue`, `currentValue`, and `index` from. That you're iterating over the array in here means you're iterating over the array while iterating over the array.

lib/assessment.js
```diff
@@ -8,15 +8,45 @@
...
+    // set the updated result
+    result = currentValue < previousValue ? currentValue : previousValue;
```
srhoulam: The condition segment of this ternary is incorrect for the approach you've taken here, but correct in the desired callback function.

lib/assessment.js
```diff
@@ -8,15 +8,45 @@
...
+  }
+  // return the value or it won't pass grunt test
+  return result;
```
srhoulam: So, overall, this callback works, but it does so in a very wasteful manner. It checks `currentValue < previousValue` and assigns the lesser of the two to `result`, an operation that has the same result every time if the values it depends on are the same, `array.length` times.

3/4

resource11: For a beginner programmer, if a function works *and* passes a grunt test, I don't see a reason for deduction points for wastefulness.

lib/assessment.js
```diff
@@ -8,15 +8,45 @@
...
+//   }
+// };
+
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
 
+// cool! Looks like we don't need to pass index and array
+// into the callback function?
```
srhoulam: And you don't necessarily have to use everything that's passed in.

lib/assessment.js
```diff
@@ -8,15 +8,45 @@
...
+var sum = function sum(previousValue, currentValue) {
+  return previousValue + currentValue;
```
srhoulam: Great. 1

lib/assessment.js
```diff
@@ -24,22 +54,57 @@ var sum = function sum() {
...
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
-
+  // set initial result value
+  var result = false;
+  array.forEach(function(currentValue, index, array){
+    // this means if the result of invoking callback
+    // returns a truthy value...
+    if (callback(currentValue)) {
+      result = true;
```
srhoulam: This also works. Good alternative solution.

lib/assessment.js
```diff
@@ -24,22 +54,57 @@ var sum = function sum() {
...
+    }
+  });
+  return result;
```
srhoulam: Well done. 1

lib/assessment.js
```diff
@@ -24,22 +54,57 @@ var sum = function sum() {
...
+// callback(currentValue) is truthy, add that currentValue
+// to the new array
+var select = function select(array, callback, thisArg) {
+  var newArray = [];
+  array.forEach(function(currentValue, index, array) {
+    if (callback(currentValue)) {
+      newArray.push(currentValue);
+    }
+  });
+  return newArray;
```
srhoulam: Correct. 1

lib/assessment.js
```diff
@@ -24,22 +54,57 @@ var sum = function sum() {
...
+// if callback() returns something truthy, immediately return that
+// element. That element is an array index value.
+var find = function find(array, callback, thisArg) {
+  var result = null;
+  array.forEach(function(currentValue, index, array){
+    if (callback(currentValue, index, array)) {
+      result = currentValue;
+    }
+  });
+  return result;
```
srhoulam: That's it. 1

srhoulam:
4.75/5

resource11:
Sweet! Thanks for taking the time to explain the finer points. Really liked seeing that cool way of exiting a for loop you showed last Friday.

PR #42, opened 2015-10-30
fixed assessment

sdavidson140:


lib/assessment.js
```diff
@@ -8,40 +8,64 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-var min = function min() {
-
+var min = function min(previousValue, currentValue) {
+  return previousValue < currentValue? previousValue : currentValue;
```
srhoulam: Yes. 1

lib/assessment.js
```diff
@@ -8,40 +8,64 @@
...
 };
 
+
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-var sum = function sum() {
-
+var sum = function sum(previousValue, currentValue) {
+  return previousValue += currentValue;
```
srhoulam: This will work correctly, but assigning to a parameter is frowned upon except when setting a default value or in rare cases where it's the way to go. Returning `previousValue + currentValue` was an ideal solution; this works because the two lines evaluate to the same thing, and so the return value is the same.

1

lib/assessment.js
```diff
@@ -8,40 +8,64 @@
...
 // Write a function that returns true
 // if invoking the callback for any element
 // in the array returns a truthy value
 var any = function any(array, callback, thisArg) {
-
+  var result = false;
+  array.forEach(function(currentValue, index, array){
+    result = result || callback(currentValue, index, array);
+  });
+  return result;
```
srhoulam: Yep. 1

lib/assessment.js
```diff
@@ -8,40 +8,64 @@
...
 // the callback returns a truthy value
 var select = function select(array, callback, thisArg) {
-
+  var newArray = [];
+  array.forEach(function(currentValue, index, array){
+    if (callback(currentValue, index, array)) {
+      newArray.push(currentValue);
+    }
+  });
+  return newArray;
```
srhoulam: Right on. 1

lib/assessment.js
```diff
@@ -8,40 +8,64 @@
...
+//  }
+//};
+var find = function find(array, callback, thisArg){
+  var firstTruthy;
+  array.forEach(function(currentValue, index, array){
+    if (!firstTruthy && callback(currentValue, index, array)){
+      firstTruthy = currentValue;
+    };
+  });
+  return firstTruthy;
```
srhoulam: Good. 1

srhoulam:
5/5

PR #47, opened 2016-01-26
A.Jordon diag response

ajordon:


lib/diagnostic.js
```diff
@@ -9,14 +9,26 @@
...
 // the minimum value in the array
 const min = function min() {
+  let answer = arguments[0];
+  for (let i = 1; i < arguments.length; i++) {
+    if (answer > arguments[i]) {
+      answer = arguments[i];
+    }
+  }
 
+  return answer;
```
MattB-GA-Boston: We were actually looking for you to write a callback function that you could pass into reduce as an argument; in other words,
```
[1,2,3].reduce(min)
// => 1
```

lib/diagnostic.js
```diff
@@ -9,14 +9,26 @@
...
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 const sum = function sum() {
+  let answer = arguments[0];
+  for (let i = 1; i < arguments.length; i++) {
+    answer += arguments[i];
+  }
 
+  return answer;
```

lib/diagnostic.js
```diff
@@ -24,7 +36,13 @@ const sum = function sum() {
 // if invoking the callback for any element
 // in the array returns a truthy value
 const any = function any(array, callback) {
+  for (let i =0; i < array.length; i++) {
+    if (!callback(array[i], i, array)) {
+      return false;
+    }
+  }
 
+  return true;
```
MattB-GA-Boston: Close, but not quite. As written, this code will return `false` if any elements don't pass the condition. If we were to call `any` on the array `[1,3,6]`, and we had a callback that returned `true` is a number were even, this function would read the first element (`1`) and when the test came back false it would return `false` from the function. Clearly, though, this array contains an even element.

lib/diagnostic.js
```diff
@@ -32,14 +50,27 @@ const any = function any(array, callback) {
...
 // the callback returns a truthy value
 const select = function select(array, callback) {
+  let newArray = new Array();
+  for (let i =0; i < array.length; i++) {
+    if (callback(array[i], i, array)) {
+      newArray.push(array[i]);
+    }
+  }
 
+  return newArray;
```
MattB-GA-Boston: Correct.

lib/diagnostic.js
```diff
@@ -32,14 +50,27 @@ const any = function any(array, callback) {
...
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
 const first = function first(array, callback) {
+  for (let i =0; i < array.length; i++) {
+    if (callback(array[i], i, array)) {
+      return array[i];
+    }
+  }
 
+  return false;
```
MattB-GA-Boston: Yep, this works.

MattB-GA-Boston:
2.5 / 5 (-1 on Q1 and Q2; -0.5 on Q3)

Let me know if you have any questions, or want further feedback!

PR #48, opened 2016-01-26
Finish diagnostic

dumplingpuff:


lib/diagnostic.js
```diff
@@ -8,15 +8,18 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-const min = function min() {
-
+const min = function min(array) {
+  array.reduce(function(prev, curr) {
+  });
```
i.e.
```
[1,2,3].reduce(min);
// => 1
```

lib/diagnostic.js
```diff
@@ -8,15 +8,18 @@
...
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-const sum = function sum() {
-
+const sum = function sum(array) {
+  array.reduce(function(prev, curr) {
+    return prev + curr;
+  });
```

lib/diagnostic.js
```diff
@@ -32,14 +40,26 @@ const any = function any(array, callback) {
 // made up of the array elements for which invoking
 // the callback returns a truthy value
 const select = function select(array, callback) {
-
+  let newArray = [];
+  for (let i = 0; i < array.length; i++) {
+    if (callback(array[i])) {
+      newArray.push(callback(array[i]));
+      return newArray;
```
MattB-GA-Boston: Whoops... we don't want to put the return statement there. If we do, it will return the new array after adding the first element - we will never get back more than the first matching result.

MattB-GA-Boston:
2.5 / 5 (-1 on Q1 and Q2; -0.5 on Q4)

PR #49, opened 2016-01-26
partially completed

byau:


lib/diagnostic.js
```diff
@@ -9,14 +9,24 @@
...
 // the minimum value in the array
 const min = function min() {
-
+  let smallest = arguments[0];
+  for (let i = 1; i < arguments.length; i++) {
+    if (smallest > arguments[i]) {
+      smallest = arguments[i];
+    }
+  }
+  return smallest;
```
i.e.
```
[1,2,3].reduce(min);
// => 1
```

byau: yea i know.  i msg jason on slack that all my answer would be wrong.

lib/diagnostic.js
```diff
@@ -9,14 +9,24 @@
...
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 const sum = function sum() {
-
+  let total = arguments[0];
+  for (let i = 1; i < arguments.length; i++) {
+    total += arguments[i];
+  }
+  return total;
```

lib/diagnostic.js
```diff
@@ -24,7 +34,13 @@ const sum = function sum() {
...
 // in the array returns a truthy value
 const any = function any(array, callback) {
-
+  for (let i = 0; i < array.length; i++) {
+    if(callback(array[i])){
+      return true;
+    } else {
+      return false;
+    }
+  }
```
MattB-GA-Boston: Almost. This part is correct:
```
if(callback(array[i])){
     return true;
}
```
But the `else` breaks it. Can you see why?

lib/diagnostic.js
```diff
@@ -32,14 +48,24 @@ const any = function any(array, callback) {
...
 // the callback returns a truthy value
 const select = function select(array, callback) {
-
+  let newArray = [];
+  for (let i = 0; i < array.length; i++) {
+    if(callback(array[i])){
+      newArray.push(array[i]);
+    }
+  }
+  return newArray;
```
MattB-GA-Boston: Correct.

lib/diagnostic.js
byau: coz it will exit the loop the minute it hit false.  so it will not be able to go thru the whole array.

lib/diagnostic.js
```diff
@@ -32,14 +48,24 @@ const any = function any(array, callback) {
...
 // Question 5
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
 const first = function first(array, callback) {
-
+  for (let i = 0; i < array.length; i++) {
+    if(callback(array[i])){
+      return array[i];
+    }
+  }
```
MattB-GA-Boston: What if the callback never returns a truthy value?

MattB-GA-Boston:
2.25 / 5 (-1 on Q1 and Q2; -0.5 on Q3; -0.25 on Q5)

PR #50, opened 2016-01-26
Diagnostic response

MFBrewster:


lib/diagnostic.js
```diff
@@ -8,15 +8,15 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-const min = function min() {
-
+const min = function min(prev, curr, index, array) {
+  return curr < prev ? curr : prev;
```
MattB-GA-Boston: Yep!

lib/diagnostic.js
```diff
@@ -8,15 +8,15 @@
...
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-const sum = function sum() {
-
+const sum = function sum(prev, curr, index, array) {
+  return prev + curr;
```
MattB-GA-Boston: Exactly.

lib/diagnostic.js
```diff
@@ -24,7 +24,12 @@ const sum = function sum() {
 // if invoking the callback for any element
 // in the array returns a truthy value
 const any = function any(array, callback) {
-
+  for (let i = 0; i < array.length; i++)  {
+    if (callback(array[i], i, array) === true)  {
```
MattB-GA-Boston: It's not necessary to use ` === true` here; if the callback is returning a truthy value, that's enough to tell the `if` which way to go.

lib/diagnostic.js
```diff
@@ -32,14 +37,25 @@ const any = function any(array, callback) {
 // made up of the array elements for which invoking
 // the callback returns a truthy value
 const select = function select(array, callback) {
-
+let newArray = new Array(0);
+  for (let i = 0; i < array.length; i++)  {
+    if (callback(array[i], i, array) === true)  {
```
MattB-GA-Boston: See previous note.

lib/diagnostic.js
```diff
@@ -32,14 +37,25 @@ const any = function any(array, callback) {
...
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
 const first = function first(array, callback) {
-
+  for (let i = 0; i < array.length; i++)  {
+    if (callback(array[i], i, array) === true)  {
+      return array[i];
+    }
+  }
+  return undefined;
```
MattB-GA-Boston: I would say that you should return `null` rather than `undefined`. You would also get `undefined` if your function didn't return anything at all; choosing `null `makes it clear that you are returning a value (even if that value is 'no results').

MattB-GA-Boston:
5 / 5

Very nice work. The notes above are about code quality; while what you wrote isn't "wrong", it could still be tighter.

PR #51, opened 2016-01-26
not even close, need to go over this

nwilcox11:


lib/diagnostic.js
```diff
@@ -9,15 +9,40 @@
...
+        args = arguments;
+    }
+    let answer = args[0];
 
+    for(let i = 1; i < args.length; i++) {
+        if(answer > args[i]) {
+            answer += args[i];
+        }
+        return answer;
+    }
```
i.e.
```
[1,2,3].reduce(min);
// => 1
```

lib/diagnostic.js
```diff
@@ -9,15 +9,40 @@
...
 
+    for(let i = 1; i < array.length; i++) {
+        answer += array[i];
+    }
+    return answer;
 };
+let array = [2,4,6];
+array.reduce(function(a, b) {
+    return a + b;
+});
```
MattB-GA-Boston: I think you sort of arrived at the right answer eventually.
```
const sum = function(a,b) {
  return a + b;
}
```
When you run `[2,4,6].reduce(sum)` you should get a result of 12

MattB-GA-Boston:
1 / 5

I'm not sure you're necessarily so lost on this material - it seems like you just didn't understand the questions and got stuck. I strongly recommend you give this one another shot.

PR #52, opened 2016-01-26
Array Method Diagnostic

msyao-github:


lib/diagnostic.js
```diff
@@ -8,17 +8,23 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-const min = function min() {
-
+const min = function min(array,previousValue, currentValue) {
+  return (previousValue <currentValue ? previousValue : currentValue);
```
MattB-GA-Boston: Correct!

lib/diagnostic.js
```diff
@@ -8,17 +8,23 @@
...
 
+
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-const sum = function sum() {
+const sum = function sum(array) {
+  array.reduce(function(previousValue,currentValue){
+    return a+b;
+  })
```
MattB-GA-Boston: We were specifically looking for the callback here.
```
const sum = function(a,b) {
  return a + b;
};
```

lib/diagnostic.js
```diff
@@ -38,10 +45,21 @@ const select = function select(array, callback) {
...
-const first = function first(array, callback) {
 
+
+const first = function (array, callback) {
+    if (typeof array[i] !=== 'undefined') {
+        return undefined;
+    }
+
+    for (let i = 0; i < arr.length; i++) {
+        if (i in this. && callback (previousValue,this[i],a)) return this[i];
```
MattB-GA-Boston: The `i in this` bit is unnecessary (and I'm not sure it's valid syntax). The callback also needs only to operate on any given element.
i.e.
```
for (let i = 0; i < arr.length; i++) {
  if (callback(array[i])) {
    return array[i];
  }
}
```

lib/diagnostic.js
```diff
@@ -38,10 +45,21 @@ const select = function select(array, callback) {
...
+    }
+
+    return array[i];
```
MattB-GA-Boston: `i` is not defined outside of the loop. Additionally, if no matching elements can be found, we should probably return `null`.

MattB-GA-Boston:
2 / 5 (-0.5 on Q2; -1 on Q3 and Q4; -0.5 on Q5)

PR #53, opened 2016-01-26
answers to diagnostic

kpie89:
Def need more review


J-Weeks:
Didn't pass tests. I would recommend doing some code wars with array method for extra help.

PR #54, opened 2016-01-26
array methods diagnostic

apalmer0:


J-Weeks:
2/5. Weird. array.reduce is def a function.  You have a bunch of syntax errors though.  Be sure to run grunt test, grunt nag and node <filename>.  Looks like you got hung up on a question and didn't get the the rest.  Next time. I think you have an understand of this but just didn't do well on this diagnostic.

PR #58, opened 2016-01-26
js-array-method

kenvng:
I have no clue of what I am doing.


J-Weeks:
I'm a bit worried. Please do more coding.  code wars or node school.  I think you need more practice coding.

kenvng:
my problem is that i didn't understand it to write it.  Can you sit down
with me after class and run over function and array to help me with the
concept and better understand the materials?
Thank you,
Ken

PR #59, opened 2016-01-26
wrote some answers

Knguyen21:


J-Weeks:
3/5.  Sum and find failed tests.  Was this a matter of time or need a bit more work with array methods?

Knguyen21:
Antony and I talked about find. It's the wrong name in the test and specs. It's first in diagnostic and find in specs and test. Antony made a mistake creating it. 

As for Sum, I need more time and also need to understand reduce better.

PR #62, opened 2016-01-26
Couldn't answer any of these

kylekatz:


J-Weeks:
Check the solution branch for the correct answers.  During the workshop tomorrow do some work on arrays and I'll be around to help answer questions.  We can step through some of these.

PR #68, opened 2016-03-29
Responds to JS Array Methods Diagnostic

OllieBuilds:


lib/diagnostic.js
```diff
@@ -3,33 +3,48 @@
...
 // Question 4
 // Write a function that returns true
 // if invoking the callback for any element
 // in the array returns a truthy value
-const any = function any(array, callback) {
 
-};
+let arr = [1, 2, 3, 4];
+
+arr.some(a => a > 0);
```
RDegnen: Looking for something more like this 
```
  let testPass = false;
      testPass = true;
    }
  }
  return testPass;
};
```

RDegnen:
5 of 6 nice!

PR #70, opened 2016-03-29
answered the questions

pvdri:
didn't have all code from yesterday...

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -4,21 +4,62 @@
 // Describe in your own words what a `callback`
 // is in relation to array methods
 
+A callback is a (1) function within a (2) function that calls the (2) function.
+
+With regards to array methods, it helps you to create and populate an array.
+
+
 // Question 2
 // Describe in your own words the differences between a `for` loop
 // and the `forEach` array method
+
+A for loop just call each element of the array.  A forEach function actually operates on each element of the array.
 
 
 // Question 3
 // Implement the `reduce` array method, using arrow function
 // syntax, to return the total sum of all values in the original array
 
+let cb = function(i)
+
+{for (let i ==1; i <= array.length; i ++)
+  let result (arr[0])
+  {
+    result = result + arr[i]
+  }
+  return result;
+};
+  {
+
+  return ++1;}
+
+let map = function(array, callback) {
+  let newArr = [];
+  for (let i ==1; i <= array.length; i ++)
+  {
+    newArr = callback(array[i]);
+  }
+  return newArr;
+};
+
+console.log(map(arr, cb));
+
+module.exports = map;
+
 
 // Question 4
 // Write a function that returns true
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -4,21 +4,62 @@
...
+
+With regards to array methods, it helps you to create and populate an array.
+
+
 // Question 2
 // Describe in your own words the differences between a `for` loop
 // and the `forEach` array method
 
+A for loop just call each element of the array.  A forEach function actually operates on each element of the array.
+
```
RDegnen: A for loop can also operate on elements in an array. 
```
`for` loops can:

- can be broken out of

`forEach`

- cannot be broken out of
```

lib/diagnostic.js
```diff
@@ -4,21 +4,62 @@
...
+  {
+    newArr = callback(array[i]);
+  }
+  return newArr;
+};
+
+console.log(map(arr, cb));
+
+module.exports = map;
+
```
RDegnen: looking for this ```arr.reduce((a, b) => a + b);```

lib/diagnostic.js
```diff
@@ -4,21 +4,62 @@
...
 
 // Question 4
 // Write a function that returns true
 // if invoking the callback for any element
 // in the array returns a truthy value
+
+
 const any = function any(array, callback) {
+   for (let i === 0, i < array.length; i ++)
```
RDegnen: you are comparing i to 0 here instead of setting i to equal zero

lib/diagnostic.js
```diff
@@ -4,21 +4,62 @@
...
+    if callback(a[i])  = true {
```
RDegnen: a will be undefined and your missing parenthesis around your if statement condition

lib/diagnostic.js
```diff
@@ -4,21 +4,62 @@
...
+      return true;
+    }
+    }
+
```
```
  let testPass = false;
      testPass = true;
    }
  }
  return testPass;
};
```

lib/diagnostic.js
```diff
@@ -26,10 +67,26 @@ const any = function any(array, callback) {
...
 // the callback returns a truthy value
+
+let newArr = [];
+
 const select = function select(array, callback) {
+  for (let i === 0, i < array.length; i ++)
+   if callback(a[i])  = true {
+     return newArr;
+   }
+   }
```
```
  let passingElements = [];
      passingElements.push(array[i]);
    }
  }
  return passingElements;
};
```

lib/diagnostic.js
```diff
@@ -4,21 +4,62 @@
 // Describe in your own words what a `callback`
 // is in relation to array methods
 
+A callback is a (1) function within a (2) function that calls the (2) function.
```
RDegnen: A callback is a function that is called from within another function

PR #71, opened 2016-03-29
Complete js-array-methods diagnostic

philschoof:


lib/diagnostic.js
```diff
@@ -27,9 +35,29 @@ const any = function any(array, callback) {
...
+  let testPass = true;
+  for (let i = 0; i < arr.length; i++){
+    if ( typeof arr[i] !== 'number'){
+      testPass = false;
+    }
+  }
+  return testPass;
+};
+
+arr.every(isNum);
```
RDegnen: not bad, but you also could have done ```arr.every(a => typeof a === 'number');```

RDegnen:
6 of 6 yay

PR #73, opened 2016-03-29
answers to js-array-methods

V4Vince:


lib/diagnostic.js
```diff
@@ -4,32 +4,68 @@
 // Describe in your own words what a `callback`
 // is in relation to array methods
 
+They are basically a function that store value to be passed inside another function
+
```
RDegnen: A callback is a function passed to an array method to do whatever is specified in the callback on each array element

lib/diagnostic.js
```diff
@@ -4,32 +4,68 @@
...
+let evenOrNot = function(num){
+  return num % 2 === 0;
+}
 
+const any = function any(array, callback) {
+  let testPast = true;
+  for (let i = 0; i < array.length; i++) {
+    if (callback(array[i]) === false) {
+      testPast = false;
+      break;
```
RDegnen: you dont need this break

PR #75, opened 2016-03-29
Response to questions

bryansarnold:


lib/diagnostic.js
```diff
@@ -3,33 +3,54 @@
 // Question 1
 // Describe in your own words what a `callback`
 // is in relation to array methods
-
+You are accessing an array that has already been stored in a variable beforehand.
```
RDegnen: A callback is a function passed to an array method to determine the action put upon each element of an array being iterated over.

lib/diagnostic.js
```diff
@@ -3,33 +3,54 @@
...
 // if invoking the callback for any element
 // in the array returns a truthy value
 const any = function any(array, callback) {
-
-};
+  let testTrue = true;
+    for (let i = 0; i < array.length; i++) {
+      if (callback(array[i])) ==== true) {
+        testTrue = false;
+        break;
```
RDegnen: you dont need the break

lib/diagnostic.js
```diff
@@ -3,33 +3,54 @@
...
+map(arr, cb);
 };
 
 // Question 6
 // Implement the `every` function to test if all elements
 // in an array are a `typeof` number
+
+let testArray = [1, 8, 60, 122];
+
+let testPass = testArray.every(a => typeof a === 10);
```
RDegnen: instead of 10 it should be `'number'`

RDegnen:
4.5 of 6 nice!

PR #76, opened 2016-03-29
Response to Diagnostic

rdelvallej32:


lib/diagnostic.js
```diff
@@ -3,33 +3,87 @@
...
+ };
+
 const select = function select(array, callback) {
+  let newArray = [];
+
+
+  for (let i = 0; i < array.length; i++) {
+    if(array[i] === true) {
+    newArray[i] = callback(array[i]);
+  }
```
RDegnen: instead of ```
if(array[i] === true) {
 +    newArray[i] = callback(array[i]);
 +  }
```
it should be 
```
      newArray[i] = array[i];
    }
```

RDegnen:
5 out of 6 nice!

PR #77, opened 2016-03-29
Add answers to diagnostic.

Jcornmanhomonoff:


lib/diagnostic.js
```diff
@@ -3,33 +3,65 @@
 // Question 1
 // Describe in your own words what a `callback`
 // is in relation to array methods
+A callback is a function that gets passed through another function as an argument.
 
 // Question 2
 // Describe in your own words the differences between a `for` loop
 // and the `forEach` array method
-
+A for loop runs through an array x amount of times and a the forEach array method runs through each individual index in an arary.
```
```
for` loops can:

- can be broken out of

`forEach`

- cannot be broken out of
```

lib/diagnostic.js
```diff
@@ -3,33 +3,65 @@
...
+let arr = [1, 2, 3];
+
+let cb = function(num){
+  return num % 2 === 0;
+}
 
+let map = function(arr, cb){
+  let newArr = [];
+  for (let i = 0; i < arr.length; i++) {
+    if (cb(ar[i]) === false){
```
RDegnen: you would check for true here not false, and then push that element into the new array

RDegnen:
3 of 6 getting better!

PR #80, opened 2016-03-29
Complete function method diagnostic

cuprous:


cuprous:
The grunt tests lied to me.

RDegnen:
6 out of 6 high-five

PR #81, opened 2016-03-29
Submit js-array-methods-diagnostic

MichaelB167:
I was pretty lost on how to approach these and will seek help in a 1-1 Wednesday if I don't make significant progress on this in the meantime.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -4,22 +4,40 @@
 // Describe in your own words what a `callback`
 // is in relation to array methods
 
+A callback function affects what is going in another function or array method
+by telling you what to do with every method that you're iterating over.
+
 // Question 2
 // Describe in your own words the differences between a `for` loop
 // and the `forEach` array method
 
+A 'for' loop iterates over an array whereas the 'forEach' array method performs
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -4,22 +4,40 @@
 // Describe in your own words what a `callback`
 // is in relation to array methods
 
+A callback function affects what is going in another function or array method
+by telling you what to do with every method that you're iterating over.
```
RDegnen: Ya close. A callback is just a function that is called by another function. It can be called on each array element to do something to them

lib/diagnostic.js
```diff
@@ -4,22 +4,40 @@
...
+
 // Question 2
 // Describe in your own words the differences between a `for` loop
 // and the `forEach` array method
 
+A 'for' loop iterates over an array whereas the 'forEach' array method performs
+a callback function on an array.
```
```
`for` loops can:

- can be broken out of

`forEach`

- cannot be broken out of
```

RDegnen:
Ya definitley set up a 1 on 1 with one of us to go over this

PR #82, opened 2016-03-29
Answer Array Methods Questions

susandaiute:


lib/diagnostic.js
```diff
@@ -3,33 +3,51 @@
...
 // Write a function that returns true
 // if invoking the callback for any element
 // in the array returns a truthy value
-const any = function any(array, callback) {
 
-};
+let testArray = [8,7];
+
+let testPass = testArray.every(a => a % 4 === 0);
+console.log(testPass);
```
```
  let testPass = false;
      testPass = true;
    }
  }
  return testPass;
};
```

lib/diagnostic.js
```diff
@@ -3,33 +3,51 @@
...
 // made up of the array elements for which invoking
 // the callback returns a truthy value
-const select = function select(array, callback) {
 
-};
+let testArray = [4,7,82,64,24,100];
+
+let testPass = testArray.filter(a => a % 4 === 0);
+console.log(testPass);
+
```
```
  let passingElements = [];
      passingElements.push(array[i]);
    }
  }
  return passingElements;
};
```

RDegnen:
4 out of 6 not bad!

PR #91, opened 2016-07-19
Add diagnosic.md

NatasaPeic:
* Didn't finish question 1 and 5.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,39 +8,128 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-const min = function min() {
+
+let arr = [1,2,3];
+
+
+let callbackMin = function (element) {
+
+  let firstElement = arr[0];
+
+  for (let i = 0; i < array.length; i++) {
+    let result =   array[i];
+
+if (firstElement < result) {
+  return firstElement;
+}
+
+  }
+
+};
+
+const min = function min(array, callback) {
+
 
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-const sum = function sum() {
+const sum = function sum(prev, curr) {
+
+  return prev + sum;
 
 };
+
+const reduce = function(arr, callback, init) {
+
+  let result;
+
+  if(init!=undefined){
+    result = init;
+  }
+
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,40 +8,129 @@
...
+if (firstElement < result) {
+  return firstElement;
+}
+
+  }
+
+};
+
+const min = function min(array, callback) {
+
```
laurenfazah: You're on the right track with this. Remember callbacks with `reduce` are dealing with previous and current values. Here is one solution:

```
};
```

lib/diagnostic.js
```diff
@@ -8,40 +8,129 @@
...
+reduce(arr, sum, 0);
+
+
+
 // Question 3
 // Write a function that returns true
 // if invoking the callback for any element
 // in the array returns a truthy value
+
```
laurenfazah: So for question 3, you'd just want to loop over an array and if any elements applied to a callback return `true`, the function can return `true`.

```
    if (callback(array[i])) {
      return true;
    }
  }

  return false;
};
```

lib/diagnostic.js
```diff
@@ -8,40 +8,129 @@
...
 const select = function select(array, callback) {
 
+  let result =[];
+
+  for (let i = 0; i < array.length; i++) {
+    result.push(callback(array[i]));
+  }
+return result;
 };
```
laurenfazah: Close! Here, you'd only want to push into `result` if `callback(array[i])` evaluates to true.

```
  
   let result =[];
 
      if  (callback(array[i])){
         result.push(callback(array[i]));
      }
   }
 return result;
 };
```

lib/diagnostic.js
```diff
@@ -8,40 +8,129 @@
...
+select(arr, thruthyVal);
+
+
 // Question 5
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
+
+
+
 const first = function first(array, callback) {
```
laurenfazah:
```
const first = function(array, callback){
      if  (callback(array[i])){
         return array[i];
      }
   }
}
```

laurenfazah:
Great effort on this, Natasha. I've corrected your answers and provided solutions. Let me know if you have any questions or concerns relating to it.

PR #92, opened 2016-07-19
Add answer to question

jlmitch1989:
I was only able to figure out question 2.  Need more help.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -15,8 +15,8 @@ const min = function min() {
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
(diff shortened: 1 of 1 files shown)


laurenfazah:
Hey Josh - great work on Q2 :). I'm pasting solutions to the rest of the diagnostic below. Let me know if you'd like to talk more about this after looking it over some. Keep up the effort!

```
// Question 1
};

// Question 3
      return true;
    }
  }

  return false;
};

// Question 4
  let result = [];
      result.push(currentValue);
    }
  });

  return result;
};

// Question 5
  let result;
      result = currentValue;
    }
  });

  return result;
};
```

PR #93, opened 2016-07-19
Complete JS Array Methods Diagnostic

gabescarbrough:
Answered questions about JS Array Methods. Confident in my answers for questions
1 and 2. I think I was on the right track for 3 and 4, but I don't think I
understand fat arrow functions, so they don't work. Didn't have time to start 5.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,15 +8,19 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-const min = function min() {
-
+const min = function min(prev, curr) {
+  if (curr < prev){
+    return curr;
+  } else {
+    return prev;
+  }
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-const sum = function sum() {
-
+const sum = function sum(prev, curr) {
+  return prev + curr;
 };
 
 // Question 3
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,31 +8,35 @@
...
 };
 
 // Question 3
 // Write a function that returns true
 // if invoking the callback for any element
 // in the array returns a truthy value
 const any = function any(array, callback) {
-
+  return array.any(element => callback(currentValue, index, array));
 };
```
laurenfazah: Your thinking was on the right track, but this was just meant to be a function iterating over an array.

```
      return true;
    }
  }

  return false;
};
```

lib/diagnostic.js
```diff
@@ -8,31 +8,35 @@
...
 
 // Question 4
 // Write a function that returns a new array
 // made up of the array elements for which invoking
 // the callback returns a truthy value
 const select = function select(array, callback) {
-
```
laurenfazah:
```

// Question 4
  let result = [];
      result.push(currentValue);
    }
  });

  return result;
};

// Question 5
  let result;
      result = currentValue;
    }
  });

  return result;
};
```

laurenfazah:
Good effort here, Gabe. I added the solutions in for 3-5 in case you wanted to revisit them.

PR #95, opened 2016-07-19
Response

jordanallain:


lib/diagnostic.js
```diff
@@ -8,38 +8,57 @@
...
 // if invoking the callback for any element
 // in the array returns a truthy value
 const any = function any(array, callback) {
-
+  for (let i = 0; i < array.length; i++) {
+    if(array[i]){
+      return true;
+    }
+  }
 };
```
laurenfazah: So close here. You'd just need to return `false` from the function in the event nothing is true.

lib/diagnostic.js
```diff
@@ -8,38 +8,57 @@
...
 // the callback returns a truthy value
 const select = function select(array, callback) {
-
+  let array2 = [];
+  for (let i = 0; i < array.length; i++) {
+    if(callback(array[i])){
+      array2.push(array[i]);
+    }
+  }
 };
```
laurenfazah: Similar issue here - very close, but you'd need to remember to return the array you've accumulated values into.

laurenfazah:
Excellent work, Jordan! Just a couple of missing `return` statements, but otherwise you're on exactly the right track.

PR #96, opened 2016-07-19
Answered all questions 1-5.

alex-braun:
5 passing, 3 failing.  New array issues.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,39 +8,60 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-const min = function min() {
-
+const min = function min(prev, curr) {
+  if (curr < prev) {
+    return curr;
+  }
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
-const sum = function sum() {
-
-};
+const sum = function sum(prev, curr) {
+    return prev + curr;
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,39 +8,60 @@
...
-};
+    let result = false;
+    for (let i = 0; i < array.length; i++) {
+      if (callback(array[i]) === true) {
+        let result = [];
+      result.push(callback(array[i]));
+    }
+     }
+     return result;
+   };
```
laurenfazah: So close here! If result is your accumulator and you want to be pushing elements into it, it'd be best to initialize it as an empty array. 

```
  let result = [];
      result.push(currentValue);
    }
  });

  return result;
};
```

lib/diagnostic.js
```diff
@@ -8,39 +8,60 @@
...
-
-};
+  let result = false;
+     for (let i = 0; i < array.length; i++) {
+       if (!callback(array[i])) {
+         result = true;
+       }
+     }
+     return result;
+   };
```
laurenfazah: Extremely close here - you want to return `true` if the callback returns true, thus you'd want to check for ` if (callback(array[i])) {`

laurenfazah:
Great work here, Alex! Keep up the effort.

PR #97, opened 2016-07-19
Add answer to array method diagnostics

EchoXin:


laurenfazah:
Hi Ying - I know we briefly talked about array methods yesterday, so let me know if you still have questions about them or this diagnostic. Here are the solutions for questions 2-5 (good work on Q1).

```

// Question 2
const sum = function sum(previousValue, currentValue) {
  return previousValue + currentValue;
};

// Question 3
      return true;
    }
  }

  return false;
};

// Question 4
  let result = [];
      result.push(currentValue);
    }
  });

  return result;
};

// Question 5
  let result;
      result = currentValue;
    }
  });

  return result;
};
```

PR #98, opened 2016-07-19
Worked on daily diagnostic

natGalynker:
-the grunt test showed too many errors so I could not check work
-this was the hardest diagnostic yet had to refer to notes a lot
-did not finish all the questions but I took my time and did not rush

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,25 +8,38 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in the array
-const min = function min() {
-
+const arr = [1, 6, 7];
+const min = function(element) {
+  if(element > 6){
+    return true;
+  }else{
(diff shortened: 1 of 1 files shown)


laurenfazah:
Hey Natalia - good effort on this. When working with grunt test, those tests rely heavily on the named functions we've put in place, so try to keep the variable names the same and only replace `your answer here`. That will help with tests down the road.

You've got Q2 correct. Here are the solutions for 1, 3-5. Let me know if you have questions after reviewing them and we can discuss. 

```
// Question 1
};

// Question 3
      return true;
    }
  }

  return false;
};

// Question 4
  let result = [];
      result.push(currentValue);
    }
  });

  return result;
};

// Question 5
  let result;
      result = currentValue;
    }
  });

  return result;
};
```

PR #103, opened 2016-10-25
Respond to questions in lib/diagnostic.js

kibs86:
Update lib/diagnostic.js with responses per diagnostic instructions.

Comfort: 4
Clarity: 5

I used my class notes and MDN docs to help with this diagnostic.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,14 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current, curIndex, a) {
+  let minNum = 0;
+  for (let i = 0; i < a.length; i++) {
+    if (minNum > current) {
+      minNum = current;
+    }
+  }
+  return minNum;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,23 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current, curIndex, a) {
```
neugierige: FYI, the `(/* previous, current */)` in the boilerplate code is supposed to be a hint about what arguments are being passed into this function

lib/diagnostic.js
```diff
@@ -5,17 +5,23 @@
...
+  let minNum = 0;
+  for (let i = 0; i < a.length; i++) {
+    if (minNum > current) {
+      minNum = current;
+    }
+  }
+  return minNum;
 };
```
neugierige: I think you might have misinterpreted the question. Because there will only be 2 arguments, all we need to do in this function is compare one thing with the other. Check out this solution:
```
};
```

lib/diagnostic.js
```diff
@@ -5,17 +5,23 @@
...
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (previous, current, i, a) {
+  return (previous + current);
 };
```
neugierige: Correct, but to be clear, we are only working with 2 arguments.

neugierige:
5/5. Great job, Kara!

PR #106, opened 2016-10-25
Attempt js array methods diagnostic

kiraamaa:
Need to work on/review callback functions
Some wording of questions unclear

Comfort: 2
Clarity: 3

I used Google and my class notes to help with this diagnostic.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,12 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
+  if (previous < current) {
+    return previous;
+  } else {
+    return current;
+  }
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,21 @@
...
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
+  if (previous < current) {
+    return previous;
+  } else {
+    return current;
+  }
 };
```
neugierige: Correct. Here's another solution:
```
};
```

lib/diagnostic.js
```diff
@@ -39,17 +50,35 @@ const any = function (/* array, callback */) {
...
+
+    array.forEach(function (e, i, a) {
+      if (callback(e, i, a)) {
+        result.push(array[i]);
+      }
+    });
+    return result;
+
+  };
 };
```
neugierige: Yes, but take a look at line 60. What are you actually pushing to the array? Isn't `array[i]` the same thing as...`e`? Also be careful of your `}`s. Make sure they match.

neugierige:
5/5. Great job, Kira! You did have 1 extra `}`. As Antony mentioned last week, it's fine for now, but please try to make a habit of looking at the errors from linter.

PR #107, opened 2016-10-25
Diagnostic on 10/25/16

Sphilippe7:
Did my best on diagnostic test

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,7 +5,11 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (min, callback) {
+  for (let i = 0; i < array.length; i++) {
+      callback(array[i], i, array);
+    }
+  };
 
 };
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,7 +5,11 @@
...
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (min, callback) {
+  for (let i = 0; i < array.length; i++) {
+      callback(array[i], i, array);
+    }
+  };
 
 };
```
neugierige: Some of your ideas are totally relevant. But the prompt said that you are writing a callback. In other words, `const min` is a callback (aka function) that will be used as a callback. Does that make sense? Take a look at this solution:
```
};
```

lib/diagnostic.js
```diff
@@ -14,8 +18,11 @@ const min = function (/* previous, current */) {
...
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (total, callback) {
+  for (let i = 0; i < array.reduce(sum); i++) {
+      callback(array[i], i, sum);
+    }
+  };
 };
```
neugierige: Here's 1 possible solution:
```
  return previous + current;
};
```

neugierige:
0/5. Hi Stanley, I know this was a hard one. The questions were actually referring to exactly the code you wrote in class yesterday. Please go back to that lesson and take a look, and schedule a 1:1 to review it.

PR #111, opened 2016-10-25
respond to diagnostic

ioncrash:
respond to diagnostic


lib/diagnostic.js
```diff
@@ -5,17 +5,21 @@
...
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
+  if (previous < current) {
+    return previous;
+  } else {
+    return current;
+  }
 };
```
neugierige: FYI, here's a solution using a ternary operator:
```
};
```

neugierige:
5/5. Excellent job!

PR #112, opened 2016-10-25
Complete diagnostic

Bowers96:
Having a difficult time understanding what is going on but have a vague
understanding.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,12 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
+  let previous = array[0];
+  array.slice().forEach(fucntion (e, i, a) {
+    callback(previous, e, i + 1, array);
+  });
+  return previous
 };
 
 // Question 2
@@ -14,8 +18,8 @@ const min = function (/* previous, current */) {
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (previous, current) {
+[0, 1, 2, 3, 4].reduce( (prev, curr) => prev + curr );
 };
 
 // All of the callbacks in questions 3 through 5 should
@@ -29,9 +33,10 @@ const sum = function (/* previous, current */) {
 // in the array returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.some`
-const any = function (/* array, callback */) {
-
+const any = function isBigEnough(element, index, array) {
+    return element >= 10;
 };
+[12, 54, 18, 130, 44].every(isBigEnough);
 
 // Question 4
 // Write a function that returns a new array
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,21 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
+  let previous = array[0];
```
neugierige: what is `array`?

lib/diagnostic.js
```diff
@@ -5,17 +5,21 @@
...
+  array.slice().forEach(fucntion (e, i, a) {
```
neugierige: what does `.slice` accomplish here? is it necessary?

lib/diagnostic.js
```diff
@@ -5,17 +5,21 @@
...
+    callback(previous, e, i + 1, array);
+  });
+  return previous
 };
```
neugierige: Please review this 1 possible solution. Note that it utilizes a ternary operator.
```
};
```

lib/diagnostic.js
```diff
@@ -29,29 +33,30 @@ const sum = function (/* previous, current */) {
 // in the array returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.some`
-const any = function (/* array, callback */) {
-
+const any = function isBigEnough(element, index, array) {
+    return element >= 10;
 };
+[12, 54, 18, 130, 44].every(isBigEnough);
```
neugierige: It seems that you misunderstood the question. The question is asking you to write a function that returns true if the callback returns true for any value that is passed into it. One hint: go back to the lesson and look at the `.some` method that we wrote together.

Please take a look at this solution:
```
const any = function (array, callback) {
  return !array.every((e, i, a) => !callback(e, i, a));
};
```

lib/diagnostic.js
```diff
@@ -29,29 +33,30 @@ const sum = function (/* previous, current */) {
...
 // you may use any feature of JavaScript
 // other than `Array.prototype.filter`
-const select = function (/* array, callback */) {
-
-};
-
+const select = function logArrayElements(element, index, array) {
+  console.log('a[' + index + '] = ' + element);
+}
+[2, 5, , 9].forEach(logArrayElements);
```
neugierige: Take a look at the `.filter` method you wrote in class. Here's 1 possible:
```
  const result = [];
    if (callback(e, i, a)) {
      result.push(e);
    }
  });
  return result;
```

lib/diagnostic.js
```diff
@@ -29,29 +33,30 @@ const sum = function (/* previous, current */) {
...
 // you may use any feature of JavaScript
 // other than `Array.prototype.find`
-const first = function (/* array, callback */) {
-
-};
-
+const first = function isBiggerThan10(element, index, array) {
+  return element > 10;
+}
+[12, 5, 8, 1, 4].some(isBiggerThan10); // true
```
neugierige: Here's 1 solution:
```
const first = function (array, callback) {
    if (callback(array[i])) {
      return array[i];
    }
  }
};
```

neugierige:
0/5. Hi Brandon, callbacks is a complex and advanced topic. Many people have difficulty with it. If I were you, however, I would go back to the lessons on functions and make sure you get comfortable with writing and calling functions. We will have an optional review session this afternoon. Please be sure to attend.

PR #113, opened 2016-10-25
Adds response to array-methods diagnostic

jdaly04:
Attempts to answer questions for array diagnostic

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,17 +5,22 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (previous, current) {
+    let array = [];
+      let minVal = Math.min(array[i]);
 
-};
+    }; return(min);
+
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,22 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (previous, current) {
+    let array = [];
+      let minVal = Math.min(array[i]);
+
+    }; return(min);
```
raq929: return statements must be INSIDE the function's {}
They will cause an error if they are outside.

lib/diagnostic.js
```diff
@@ -5,17 +5,22 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
```
raq929: I think you're confusing *implementing* reduce, which we did in class, and writing a callback that will be *passed* to reduce. (Which is much easier!)
Please review the reduce documentation on MDN.

raq929:
0/5
It looks like it would be a good idea for you to review this lesson.
There will be a review of functions like this during workshop today, and if you want to go over it more, please get in touch with one of us for a 1:1.

PR #114, opened 2016-10-25
Respond to diagnostic

callmelife:
Attempted diagnostic. Was quite difficult. Had a hard time with that one. 

Clarity: 1
Comfort: 1

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,17 +5,23 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (previous, current) {
+  let previous = array[0];
+    array slice(1).forEach(function (e, i, a){
+      current(previous, e, i + 1, array);
+    });
+    return previousl
+  };
 
-};
+
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (previous, current ) {
+let previous = array[0] = array.reduce(sum)
 };
 
 // All of the callbacks in questions 3 through 5 should
@@ -29,8 +35,13 @@ const sum = function (/* previous, current */) {
 // in the array returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.some`
-const any = function (/* array, callback */) {
-
+const any = function (array, callback) {
+for (let i = 0; i < array.length; i++) {
+  if (callback(array[i])) {
+    return true;
+  }
+}
+return false
 };
 
 // Question 4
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,23 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (previous, current) {
+  let previous = array[0];
```
neugierige: why are you re-declaring the variable `previous`? what is `array`?

lib/diagnostic.js
```diff
@@ -5,17 +5,23 @@
...
+    array slice(1).forEach(function (e, i, a){
+      current(previous, e, i + 1, array);
+    });
+    return previousl
+  };
+
```
neugierige: Please review this solution:
```
};
```

lib/diagnostic.js
```diff
@@ -5,17 +5,23 @@
...
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (previous, current ) {
+let previous = array[0] = array.reduce(sum)
 };
```
neugierige: Actually, this question is _much_ simpler than what you thought it was. Remember, a `callback` is a function is that is called _by_ another function.
```
  return previous + current;
```

lib/diagnostic.js
```diff
@@ -40,6 +51,7 @@ const any = function (/* array, callback */) {
 // you may use any feature of JavaScript
 // other than `Array.prototype.filter`
 const select = function (/* array, callback */) {
+let array [];
 
 };
```
neugierige: Here is 1 possible solution:
```
  const result = [];
    if (callback(e, i, a)) {
      result.push(e);
    }
  });
  return result;
```

lib/diagnostic.js
```diff
@@ -48,8 +60,8 @@ const select = function (/* array, callback */) {
 // for which invoking the callback returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.find`
-const first = function (/* array, callback */) {
-
+const first = function  array, callback) {
+  let
 };
```
neugierige: Here is 1 possible solution:
```
const first = function (array, callback) {
    if (callback(array[i])) {
      return array[i];
    }
  }
};
```

lib/diagnostic.js
```diff
@@ -29,8 +35,13 @@ const sum = function (/* previous, current */) {
...
-const any = function (/* array, callback */) {
-
+const any = function (array, callback) {
+for (let i = 0; i < array.length; i++) {
+  if (callback(array[i])) {
+    return true;
+  }
+}
+return false
 };
```
neugierige: 👍

neugierige:
1/5. Hi Matt, it seems like you are confusing callbacks with functions that have callbacks. Do you understand the difference? I would encourage you to read up on the definition of callbacks, and review the lesson from Monday.

PR #115, opened 2016-10-25
Diagnostic Tues Oct 25

RpRoderick:
Callbacks

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -1,13 +1,26 @@
 'use strict';
 
+
+//comfort:0
+//clarity: 1.7
+//Used google and class notes
+
+
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function () {                       //object min is a function (which is the minimum in the array)
+  let reduce = function (array, callback) {       //variable reduce is a function which reduces the array
+    for (i = 0, i < array.length; i++) {  //loops through the array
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -1,13 +1,26 @@
...
+
+
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function () {                       //object min is a function (which is the minimum in the array)
```
neugierige: No. `min` is a function that returns the minimum value given an array of values.

lib/diagnostic.js
```diff
@@ -1,13 +1,26 @@
...
+  let reduce = function (array, callback) {       //variable reduce is a function which reduces the array
+    for (i = 0, i < array.length; i++) {  //loops through the array
+    callback(reduce, e, array[i], a);    //uses callback to reduce to minimum #
+  });
+}
+  return array.min;
 };
```
neugierige: You have some of the right ideas here, but you are confused about your variables are, and how callbacks work. Please take a look at this solution below. Note that it uses a ternary expression, which is really just a 1-liner if statement. 
```
};
```

neugierige:
0/5. Hi Ryan, I know callbacks is a complex and advanced topic, so it's understandable that you found this one to be very difficult. However, if I were you, I would go back in the lessons to the start of the JavaScript stuff. As you have seen, all the materials build off of each other. I would go through all the diagnostics/practices again, 1 by 1, make sure I understand all the previous content.

PR #116, opened 2016-10-25
Diagnostic on array methods

sujkid:
Attempted diagnostic on array methods

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,19 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function ( previous, current )
+{
+  let result = array[0];
+  let minimum = function ( array, callback )
+  {
+    return array.forEach( e, i, a )
+    {
+      if( result[i] > array[i] )
+      {
+        resut[i] = array[i];
+      }
+    }
+  };
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,29 @@
...
+  {
+    return array.forEach( e, i, a )
+    {
+      if( result[i] > array[i] )
+      {
+        resut[i] = array[i];
+      }
+    }
+  };
 };
```
neugierige: You've misunderstood the question here. `min` is supposed to be a _callback_. Do you remember what a callback is? This means that `min` will be called by another function. Take a look at this solution:
```
};
```

neugierige: FYI, this uses a "ternary" operator, which is really just an if-statement.

lib/diagnostic.js
```diff
@@ -5,17 +5,29 @@
...
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function ( previous, current )
+{
+  let total = array.reduce( (previous, current ) => previous, current, 0;
 };
```
neugierige: `reduce` is a method available in JavaScript that takes in a callback. The question asks you to write the callback that would go _inside_ the `reduce` method.

neugierige:
0/5. Hi Suja, I know this was a hard one. It seems like you are not entirely clear on the concept of callbacks. You need to review the lesson, and read up on their definition. We will have a review session this afternoon. Please be sure to attend.

PR #117, opened 2016-10-25
Respond to JS-Array Diagnostic

rickdean:
Diadnistic responses

Comfort: 2
Clarity: 3

I used Google and my notes to help .

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,7 +5,9 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (previous, current) {
+return Math.min(previous, current);
+//previous < current ? previous : current;
 
 };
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -48,8 +51,12 @@ const select = function (/* array, callback */) {
 // for which invoking the callback returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.find`
-const first = function (/* array, callback */) {
-
+const first = function (array, callback) {
+  for (let i = 0; i < array.length; i++) {
+    if (callback === array[i]) {
```
raq929: You have a similar issue here as you did in question 3, but you're closer!
This should be:
```js
if(callback(array[i]){
  return array[i]
}
```
Callback is a function that needs to be run on each value of the array.

raq929:
2/5
It looks like you need to review callbacks and how they work. There should be a review during workshop today. I recommend trying these again after the review, and if you're still having problems with these, come see one of us for a 1:1.

PR #118, opened 2016-10-25
js-array-methods-diagnostic response 

treep78:
attempt to answer questions
For whatever reason this really isn't clicking for me. I'm going to have to spend a lot more time practicing with these functions before I really understand them.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,12 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (array)
+{
 
+  let minimum = array.reduce((p, e, i, a) => p+c);
+  console.log("Result = "+minimum);
+  return minimum;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,16 +5,21 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (array)
+{
 
+  let minimum = array.reduce((p, e, i, a) => p+c);
```
raq929: I think you're not quite understanding the question here. `min` is supposed to be a function that takes previous and current and is *passed* to reduce as a callback. You do not need to use reduce to answer this question.
```js
  if (previous < current) {
    return previous;
  } 
  return current;
}
```

lib/diagnostic.js
```diff
@@ -29,8 +34,9 @@ const sum = function (/* previous, current */) {
 // in the array returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.some`
-const any = function (/* array, callback */) {
-
+const any = function (array, callback)
+{
+  console.log(array.filter(callback(array)));
```
raq929: I'm not quite sure what you're trying to do this.

raq929:
0/5
It looks like you need to review this lesson, especially callbacks and how they work. There will be a review during workshop today. I recommend trying these again after the review, and if you're still having trouble, please find one of us for a 1:1.

PR #119, opened 2016-10-25
js-array-method Diagnostic Response

kmg1434:
No idea how to use callbacks, and I don't undestand
the purpose of the reduce method. I need some 1 on 1 help


raq929:
2/5
Your first two responses look great. It looks like you might be having problems translating the questions into code on the last three? I would review this lesson and try the questions again. If you're still having problems figuring out how to start the last three questions, please contact one of us for a 1:1.

PR #120, opened 2016-10-25
attempted 2 of diagnostic

jbuzzanco:
ran grunt test and got both wrong

comfort 0
clarity 1

used mdn and google and lessons from yesterday


raq929:
1/5

I would continue to review this lesson and the solution branch.

PR #121, opened 2016-10-25
Add response for js-array-methods-diagnostic.

coffeeisamazing:
Add response for js-array-methods-diagnostic.

Comfort: 3
Clarity: 1

I used Google and my class notes to help with this diagnostic

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,13 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
+  if (previous < current) {
+    return previous;
+  }
+  else {
+    return current;
+  }
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -29,8 +34,15 @@ const sum = function (/* previous, current */) {
...
+const any = function (array, callback) {
+  for (let i = 0; i < array.length; i++) {
+    if (callback(array[i], i, array)) {
+      return true;
+    }
+    else {
+      return undefined;
+    }
+  }
 };
```
neugierige: Two issues here: 
1) Why is your `else` clause returning `undefined` instead of `false`?
2) Where is that 2nd return statement happening? Look at its position relative to your if-statement and your for-loop.

neugierige: I think what you were trying to write is this:
```
const any = function (array, callback) {
    if (callback(array[i])) {
      return true;
    }
  }
  return false
};
```

Here's another possible solution:
```
const any = function (array, callback) {
  return !array.every((e, i, a) => !callback(e, i, a));
};
```

neugierige:
4/5. Great job, Caleb! Happy to discuss this later this afternoon.

PR #122, opened 2016-10-25
response to array methods diagnostic

kfili:
comfort 1
clarity 3

used MDN, solution, and stack overflow.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -3,11 +3,20 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in an array
-// this callback will be passed by the tests to the reduce method, e.g.
+// this callback will be passed by the tests
+// to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -3,12 +3,21 @@
 // Question 1
 // Write a callback for reduce that selects
 // the minimum value in an array
-// this callback will be passed by the tests to the reduce method, e.g.
+// this callback will be passed by the tests
+// to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
```
raq929: Looks good!

raq929:
1/5
It looks like you had the right idea for the first question. Even if you can't get to all of the questions, writing pseudo code in comments, or just comments stating what you intend to do can be helpful.
It looks like you should review this lesson and try these questions again. If you're still unclear on them, please find one of us for a 1:1.

PR #123, opened 2016-10-25
Responded to Array method diagnostic

Naki21:
Need practice.............

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,26 +5,13 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-// const minimum = function(array) {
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -29,27 +34,45 @@ const sum = function (/* previous, current */) {
...
-const any = function (/* array, callback */) {
-
+const any = function (array, callback) {
+  for (let i = 0; i < array.length; i++) {
+    if(callback(array[i]) === true){
+      return true;
+    }
+  }
+  return false;
 };
```
BenGitsCode:
```js
  const result = [];
    if (callback(e, i, a)) {
      result.push(e);
    }
  });
  return result;
};
```

BenGitsCode:
We'll be reviewing some of this material after lunch.

PR #124, opened 2016-10-25
Response

ielole:
Attempt at array methods.
comfort : 0
clarity: 1

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,7 +5,16 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -14,9 +23,12 @@ const min = function (/* previous, current */) {
...
+const sum = function (previous, current) {
+  let array = new Array(function( ){
+    for (let i = 0;)
+  };
 
-};
+});
 
 // All of the callbacks in questions 3 through 5 should
 // be called with the current element in the array,
```
BenGitsCode: Here are the solutions for one and two. Please review the solutions and run them to see what's happening. We'll be going over some of this after lunch as well 👍  

```js
// Question 1
};

// Question 2
  return previous + current;
};
```

PR #128, opened 2017-01-17
Update from node-template

gaand:
Refresh for cohort 016


bin/index.js
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env node
+'use strict';
+
+const example = require('../lib/example');
```
jrhorn424: `bin/index.js` doesn't seem necessary for this diagnostic, but I do see it in `node-template@016/master`.

grunt/nodemon.json
```diff
@@ -0,0 +1,5 @@
+{
+  "server": {
+    "script": "bin/index.js"
```
jrhorn424: Not sure what the utility is here. My thinking goes along the lines that this is like an example file, so shouldn't be included by default. It's a reminder to include it if there is a `bin` script in the material repo, no?

PR #132, opened 2017-01-19
Still working on building these.

novakw267:
I understand what these are asking for, and the definitions of them. However applying and compiling the functions in the correct way seems to be what I am unsure about. I am going to continue to study that part of it. 

comfort 3
clarity 2.5

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,14 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (previous, current) {
+  let array = previous;
+  let minimum = array.reduce(min);
+  if (typeof current === 'undefined') {
+    minimum = previous.shift();
+  }
 
+  return minimum;
 };
 
 // Question 2
@@ -14,8 +20,14 @@ const min = function (/* previous, current */) {
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
+const sum = function (previous, current) {
+  let array = previous;
+  let total = array.reduce(sum);
+    if (typeof current === 'undefined') {
+      total = previous.shift();
+    }
 
+    return total;
 };
 
 // All of the callbacks in questions 3 through 5 should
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,29 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (previous, current) {
+  let array = previous;
+  let minimum = array.reduce(min);
```
bernardlee: ```min``` is the callback function that you're writing to be passed into ```reduce```.

lib/diagnostic.js
```diff
@@ -5,17 +5,29 @@
...
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
+const sum = function (previous, current) {
+  let array = previous;
+  let total = array.reduce(sum);
```
bernardlee: ```sum``` is the callback function that you're writing to be passed into ```reduce```.

lib/diagnostic.js
```diff
@@ -39,17 +56,27 @@ const any = function (/* array, callback */) {
...
+    return true;
 };
 
 // Question 5
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.find`
-const first = function (/* array, callback */) {
+const first = function (array, callback) {
```
bernardlee: This function should return the first element that it finds, not a boolean value.

lib/diagnostic.js
```diff
@@ -39,17 +56,27 @@ const any = function (/* array, callback */) {
 // the callback returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.filter`
-const select = function (/* array, callback */) {
+const select = function (array, callback) {
```
bernardlee: This function should return an array, not a boolean value.

lib/diagnostic.js
```diff
@@ -29,8 +41,13 @@ const sum = function (/* previous, current */) {
 // in the array returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.some`
-const any = function (/* array, callback */) {
+const any = function (array, callback) {
+  let index = array.find(callback);
```
bernardlee: ```find``` returns the first value that satisfies the predicate, not an array index.

bernardlee:
0/5

PR #133, opened 2017-01-19
array methods diagnostic

laurpaik:
-comfort: 2;
-clarity: 3

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -54,8 +54,12 @@ const any = function (array, callback) {
 const select = function (array, callback) {
   for (let i = 0; i < array.length; i++) {
     let element = array[i];
-    callback(element, i, array);
+    if (callback(element, i, array)) {
+      return i;
+    }
   }
+
+  return -1;
 };
 
 // Question 5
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,24 @@
...
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (previous, current) {
+  let add = previous.reduce(previous + current);
```
bernardlee: ```previous``` can be anything and if it's not an array or something that implements ```reduce```, then it'll throw an error.

lib/diagnostic.js
```diff
@@ -39,17 +51,29 @@ const any = function (/* array, callback */) {
 // the callback returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.filter`
-const select = function (/* array, callback */) {
+const select = function (array, callback) {
```
bernardlee: This function should return an array.

bernardlee:
2/5

PR #134, opened 2017-01-19
Yen's half answers- Array-Methods-Diagnostic

yenpraja610:
Comfort: 1
Clarity: 1

:(

I used the training files from yesterday. Still not so clear...

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,15 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (originalArray) {
+  let minimum = originalArray[0];
 
+  for (let i=0; i < originalArray.length; i++) {
+    if (originalArray[i] < minimum) {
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,28 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (originalArray) {
```
bernardlee: This function should have two parameters, ```previous``` and ```current```.

lib/diagnostic.js
```diff
@@ -5,17 +5,28 @@
...
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (originalArray) {
```
bernardlee: This function should have two parameters, ```previous``` and ```current```.

bernardlee:
0/5

PR #136, opened 2017-01-19
Diagnostic -  JS array methods

sinyru:
- Comfort: 2
Clarity: 3

I used MDN for this and am not sure how to answer the last question for
the first method.


bernardlee:
4/5.  Do you mean that you don't understand what it's asking for, or how to write the code?

sinyru:
Both, not sure what its asking and how to write it.

bernardlee:
The function should traverse an array and apply the callback function to each element.  It should return the first element where applying the callback to it returns a truthy value.

PR #137, opened 2017-01-19
Response

aneela06:
Comfort: 1
Clarity: 4

I used Google and the training on array methods to help with this diagnostic.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,16 +5,25 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
+  const newArray = 0;
+  for (let i = 0; i < previous.length; i++){
+    if (previous < current){
+      return previous;
+    }
+  }
 };
+let min = array.reduce(min);
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,16 +5,25 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
+  const newArray = 0;
```
bernardlee: You're trying to declare an array here but you're setting it to be a number.

lib/diagnostic.js
```diff
@@ -5,16 +5,25 @@
...
+  for (let i = 0; i < previous.length; i++){
```
bernardlee: ```previous``` is assumed to be a number so trying to get its ```length``` property will throw an error.

lib/diagnostic.js
```diff
@@ -5,16 +5,25 @@
...
+    if (previous < current){
+      return previous;
+    }
+  }
 };
+let min = array.reduce(min);
```
bernardlee: This is in the global scope and since ```array``` doesn't exist, it will throw an error.

bernardlee:
0/5

PR #138, opened 2017-01-19
Finding solutions to questions related with callbacks

wilsonkaya:
Solve questions related with callbacks

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,14 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function ( numberArray, callback) {
+let number = numberArray[0];
+numberArray.forEach((e,i,a) => {
+  if(e < number){
+    number = e;
+  }
+});
+return number;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,27 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function ( numberArray, callback) {
```
bernardlee: The two arguments here are assumed to be numbers.

lib/diagnostic.js
```diff
@@ -5,17 +5,27 @@
...
+let number = numberArray[0];
```
bernardlee: Since the function is expecting two numbers as arguments, this will attempt to get the 0th index of a number and return ```undefined```.

lib/diagnostic.js
```diff
@@ -5,17 +5,27 @@
...
+numberArray.forEach((e,i,a) => {
```
bernardlee: This will attempt to invoke ```forEach``` on a number and throw an error.

lib/diagnostic.js
```diff
@@ -5,17 +5,27 @@
...
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (numbers, callback) {
```

bernardlee:
0/5

PR #139, opened 2017-01-19
Response

doremaxime:
Comfort: 1
Clarity: 3

Went on: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,15 +5,24 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+//The reduce() method applies a function against an accumulator and each value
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,15 +5,24 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+//The reduce() method applies a function against an accumulator and each value
+//of the array (from left-to-right) to reduce it to a single value.
 
+const min = function (previous, currentValue) {
+  array.reduce(function(accumulator, currentValue, currentIndex, array) {
```
bernardlee: ```array``` doesn't exist in this context and will cause this function to throw an error.

bernardlee:
0/5

PR #140, opened 2017-01-19
Failed Diagnostic

spw5235:
Comfort: 1
Clarity: 1

I used Google to help with this diagnostic.  Did not grasp any material from the class lesson.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,15 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (workingArray, callback) {
+  const newArray = [];
+  workingArray.reduce( function(element, index, array) {
+    if (callback(element, index, array)) {
+      newArray.push(element);
+    }
+  });
+  let minimum = array.reduce(newArray);
+  return minimum;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,26 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (workingArray, callback) {
```
bernardlee: The two function parameters are assumed to numbers in this instance.

lib/diagnostic.js
```diff
@@ -5,17 +5,26 @@
...
+  const newArray = [];
+  workingArray.reduce( function(element, index, array) {
```
bernardlee: Since the function is assumed to be called with numbers as arguments, this will throw an error because you're trying to invoke ```reduce```, which is an array method, on a number.

lib/diagnostic.js
```diff
@@ -5,17 +5,26 @@
...
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (workingArray, callback) {
```

bernardlee:
0/5

PR #141, opened 2017-01-19
Complete array method diagnostic

ddobson:
Answers added in lib/diagnostic.js

Comfort 4/5
Clarity 4/5

I used yesterday's training material as a resource.


bernardlee:
5/5

PR #142, opened 2017-01-19
Array Methods Diagnostic 1.19.2017

mario7746:
-having a lot of trouble with these. need more practice

comfort:1
clarity:3

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,7 +5,12 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (array, callback) {
+  let minimum = array.reduce();
+  for (let i = 0; i < array.length; i++) {
+  return array[i];
+  }
+   return minimum;
 
 };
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,7 +5,12 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (array, callback) {
```
bernardlee: Both parameters to this function are assumed to be numbers.  Here, you assumed that they will be an array and a function.

lib/diagnostic.js
```diff
@@ -5,7 +5,12 @@
...
+  let minimum = array.reduce();
```
bernardlee: Since ```array``` will be passed in as a number, calling ```reduce``` on it will throw an error.

lib/diagnostic.js
```diff
@@ -14,7 +19,9 @@ const min = function (/* previous, current */) {
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
+const sum = function (array, callback) {
```

bernardlee:
0/5

PR #143, opened 2017-01-19
Completes 2/5 of questions

SquirtleSquad1988:
comfort: 3.0
clarity: 3.5
Used MDN for some references

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,12 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
+  let result = previous;
+  if (current < previous) {
+    result = current;
+  }
+  return result;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -29,8 +33,13 @@ const sum = function (/* previous, current */) {
...
 // you may use any feature of JavaScript
 // other than `Array.prototype.some`
-const any = function (/* array, callback */) {
+const any = function (workingArray, callback) {
 
+    if(callback(element, index, array)) {
+      return true;
+    }
+    return false;
+});
```
bernardlee: This line and the next one break the entire script.  It should just be a single closing brace.

lib/diagnostic.js
```diff
@@ -29,8 +33,13 @@ const sum = function (/* previous, current */) {
 // in the array returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.some`
-const any = function (/* array, callback */) {
+const any = function (workingArray, callback) {
 
+    if(callback(element, index, array)) {
```
bernardlee: The logic here is correct, but you need to iterate through the array.  As it's currently written, you are trying to call callback with three arguments that don't exist.

bernardlee:
2/5

PR #144, opened 2017-01-19
Submit Diagnostic

jwaskow:
Submitting the diagnostic

Comfort: 1
Clarity: 0-1

I used my notes to help with this diagnostic.  I really don't know what we were supposed to write for the code here.  I found yesterday's morning lesson extremely difficult to follow, mainly due to the different teaching style.  I tried to refer to my notes from yesterday,  but all the variable and parameters we used are so different from the ones on this diagnostic that I really had no idea what the code is supposed to be doing.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,14 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
+  let result = 0;
+  for (let i = 0; i < previous.length; i++) {
+    if (previous[i] > result) {
+      result = previous[i];
+    }
+  }
+  return result;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,31 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
+  let result = 0;
+  for (let i = 0; i < previous.length; i++) {
```
bernardlee: ```previous``` and ```current``` are assumed to be numbers and since ```number``` doesn't have a ```length``` property, ```previous.length``` will return ```undefined``` and the loop will never run.

lib/diagnostic.js
```diff
@@ -5,17 +5,31 @@
...
+    if (previous[i] > result) {
```
bernardlee: For similar reasons above, trying to get the index of a ```number``` will return ```undefined```.

lib/diagnostic.js
```diff
@@ -5,17 +5,31 @@
...
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (previous, current, callback) {
```
bernardlee: ```callback``` doesn't need to be a parameter since the ```sum()``` itself is being used as a callback for ```reduce```.  In fact, in that particular context, ```callback``` will be treated as an array since ```reduce```'s callback parameters are the current element being iterated over, the index of the element, and the current working array.

bernardlee:
0/5

PR #146, opened 2017-01-19
Maura's diagnostic.

ladypineapple:
Comfort 1
Clarity 1

HTML CSS part of yesterday was great but the instruction in the morning really threw me for a loop. Looking forward to doing all the practice I can find over the next three days.  Thanks Obama.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,16 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
+  const workingArray = previous.slice();
+  let accumulator = current;
+  if (typeof current === 'undefined') {
+    accumulator = workingArray.shift();
+  }
+  workingArray.forEach(function (element, index) {
+    accumulator = current (previous, element, index);
+  });
+  return previous;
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,28 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
+  const workingArray = previous.slice();
```
bernardlee: ```previous``` and ```current``` are assumed to be numbers so here you're trying to invoke ```slice()``` on a number, which will throw an error.

lib/diagnostic.js
```diff
@@ -5,17 +5,28 @@
...
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (previous, current) {
+  for (let index = 0; index < previous.length; index++) {
```
bernardlee: A similar problem occurs here.  You're trying to get the ```length``` property of ```previous```, which is assumed to be a number.

bernardlee:
0/5

PR #148, opened 2017-01-19
response

paulallchin:
Clarity 4
Comfort 1

I know the  minimum += arguments[i]; is extremely wrong, at the very least the
rest of it is probably completely wrong too. I just completed the line with something
to have atom/grunt nag happy with style errors.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,6 +6,13 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (/* previous, current */) {
+  let minimum = arguments[0];
+
+  for (let i = 0; i < arguments.minimum; i++) {
+    minimum += arguments[i];
+  }
+
+  return minimum;
 
 };
 


lib/diagnostic.js
```diff
@@ -6,6 +6,13 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (/* previous, current */) {
```
bernardlee: We're using ```min()``` as a callback for ```reduce()``` so it should take at least two arguments, ```previous``` and ```current```.

lib/diagnostic.js
```diff
@@ -6,6 +6,13 @@
...
+  let minimum = arguments[0];
+
+  for (let i = 0; i < arguments.minimum; i++) {
```
bernardlee: Both arguments to ```min()``` are assumed to be numbers.  ```arguments``` doesn't have a property called ```minimum``` so ```arguments.minimum``` evaluates to ```undefined```.

bernardlee:
0/5

PR #149, opened 2017-01-19
answers to questions.

moshiko1988:
-i need help with this.
im struggling to write a funcion by mylef.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,10 +5,13 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (array, callback) {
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,10 +5,13 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (array, callback) {
+let  minimum =  array.reduce(min);
```
bernardlee: ```min()``` is the function that we are writing to pass into ```reduce()```.  We don't want to call ```reduce()``` on ```min()``` here.  Think of this function as a callback function that we're writing for somebody else to use when they invoke ```reduce()``` on an array.

bernardlee:
0/5

PR #150, opened 2017-01-19
I need to work on the last 3. I think I understood what they were ask…

alexkarasik:
…ing, but clarify

before the weekend. Also ask about the error that says your fucntions are not being defined.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -4,18 +4,33 @@
 // Write a callback for reduce that selects
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
-// `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+// `let minimum = array.reduce(min);
 
+let previousArr = [1, 2, 3, 4, 5];
+let smallestNum = previousArr[0];
+
+const min = function (previousArr, smallestNum) {
+  for(let i = 0; i < previousArr.length; i++){
+    if(previousArr[i] < smallestNum){
+      smallestNum = previousArr[i];
+    }
+  }
+  return(smallestNum);
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
-// `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -4,18 +4,33 @@
...
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
-// `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+// `let minimum = array.reduce(min);
 
+let previousArr = [1, 2, 3, 4, 5];
+let smallestNum = previousArr[0];
+
+const min = function (previousArr, smallestNum) {
```
bernardlee: This function is being passed into an invocation of ```reduce()``` as a callback function.  The three parameters for ```reduce()```'s callback are the accumulator (or previous element, as it is in this case), the current element, and the current working array (which is omitted in this case).  For this particular example, ```previous``` and ```current``` are assumed to be numbers.  Here you are assuming that the first parameter is an array when it should be a number, so trying to get the length property of the first argument in your ```for``` loop will throw an error.

lib/diagnostic.js
```diff
@@ -4,18 +4,33 @@
...
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
-// `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
+// `let total = array.reduce(sum);
+let newArr = [1, 2, 3];
+let total = 0;
 
+const sum = function (newArr, total) {
```
bernardlee: This function has the same problems with its parameters as ```min()``` does.

lib/diagnostic.js
```diff
@@ -4,18 +4,33 @@
...
+let smallestNum = previousArr[0];
+
+const min = function (previousArr, smallestNum) {
+  for(let i = 0; i < previousArr.length; i++){
+    if(previousArr[i] < smallestNum){
+      smallestNum = previousArr[i];
+    }
+  }
+  return(smallestNum);
 };
```
bernardlee: This function does pass the test, but it's because of the way the test was written.  In all other instances, it will fail.

bernardlee:
0/5

PR #151, opened 2017-01-19
Complete js-array-methods-diagnostic.

lbreguet:
-comfort: 2
-clarity: 3
-yesterday's class notes and MDN

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,15 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (callback, array) {
+  array= [];
+  callback = array.reduce(function(a, b) {
+    if (a < b) {
+      return a;
+    } else {
+      return b;
+    }
+  }, 0);
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,27 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (callback, array) {
```
bernardlee: This function is being passed into ```reduce()``` as a callback, so its arguments should be the accumulator (we named it ```previous``` in this case), ```current```, and the current working array (omitted here).

lib/diagnostic.js
```diff
@@ -5,17 +5,27 @@
...
 };
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (callback, array) {
```
bernardlee: This function has the same issues with its parameters as ```min()```.

lib/diagnostic.js
```diff
@@ -29,8 +39,11 @@ const sum = function (/* previous, current */) {
 // in the array returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.some`
-const any = function (/* array, callback */) {
-
+const any = function (array, callback) {
```
bernardlee: Good.  This works correctly.

bernardlee:
1/5

PR #152, opened 2017-01-19
diagnostic jan19

ggwilliams:
-answered diagnostic questions

Comfort:2
Clarity:2

I thought I was getting better after yesterday's lesson

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,17 +5,27 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (originalArray, callback, initalValue) {
+  const workingArray = originalArray.slice();
+  let accumulator = initalValue;
+  if (typeOf === 'undefined') {
+    accumulator = workingArray.shift();
+  }
+  workingArray.forEach(function (element, index /*, workingArray  */) {
+    accumulator = callback(accumulator, element, index, originalArray);
+  });
 };
+
+// this is not right but I think the answer is supposed to be similar to what we worked on yesterday with reduce
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (originalArray, callback, initalValue) {
+  let workingArray = originalArray.reduce(function(originalArray, callback, initalValue));
+  return
 };
 
 // All of the callbacks in questions 3 through 5 should
@@ -29,9 +39,14 @@ const sum = function (/* previous, current */) {
 // in the array returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.some`
-const any = function (/* array, callback */) {
-
+const any = function (workingArray, callback) {
+  let index = workingArray.findIndex(callback);
+  if (index < 0) {
+    return false;
+  }
+  return true;
 };
+
 
 // Question 4
 // Write a function that returns a new array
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,27 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (originalArray, callback, initalValue) {
```
bernardlee: This function is being passed into ```reduce()``` as a callback function.  Its parameters should be the accumulator (or ```previous``` element in this case), the ```current``` element, and the current working array (omitted here).  In this particular case, ```previous``` and ```current``` are assumed to be numbers.

lib/diagnostic.js
```diff
@@ -5,17 +5,27 @@
...
+  const workingArray = originalArray.slice();
```
bernardlee: Due to this function's arguments being assumed to be two numbers, ```originalArray``` is assumed to be a number and since you are trying to call ```slice()```, an array method, on a number, this will throw an error.

lib/diagnostic.js
```diff
@@ -5,17 +5,27 @@
...
+// this is not right but I think the answer is supposed to be similar to what we worked on yesterday with reduce
+
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
-
+const sum = function (originalArray, callback, initalValue) {
```
bernardlee: This function has the same parameter issues as ```min()```.

lib/diagnostic.js
```diff
@@ -5,17 +5,27 @@
...
+  let workingArray = originalArray.reduce(function(originalArray, callback, initalValue));
```
bernardlee: You have an anonymous function declaration here but no surround braces to contain its body and this will break the entire script.

lib/diagnostic.js
```diff
@@ -29,29 +39,46 @@ const sum = function (/* previous, current */) {
 // in the array returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.some`
-const any = function (/* array, callback */) {
-
+const any = function (workingArray, callback) {
```
bernardlee: Good.  This works correctly.

lib/diagnostic.js
```diff
@@ -29,29 +39,46 @@ const sum = function (/* previous, current */) {
...
 // Question 4
 // Write a function that returns a new array
 // made up of the array elements for which invoking
 // the callback returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.filter`
-const select = function (/* array, callback */) {
-
+const select = function (workingArray, callback) {
+  const newArry = new Array(workingArray.length);
```
bernardlee: The array to be returned will not necessarily have the same length as the input array.

lib/diagnostic.js
```diff
@@ -29,29 +39,46 @@ const sum = function (/* previous, current */) {
...
+  workingArray.forEach((element, index, array)=> {
+    newArray[index] = callback(element, index, array);
```
bernardlee: We only want the elements where invoking the callback function on it returns a truthy value.

lib/diagnostic.js
```diff
@@ -29,29 +39,46 @@ const sum = function (/* previous, current */) {
...
 
+//again, I think it's supposed to be similar to yesterday's lesson but I still not able to change it for the problem
+
 // Question 5
 // Write a function that returns the first element
 // for which invoking the callback returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.find`
-const first = function (/* array, callback */) {
+const first = function (workingArray, callback) {
```
bernardlee: Good.  This also works.

bernardlee:
2/5

PR #153, opened 2017-01-19
Array Methods Diagnostics

farrahj99:
Comfort: 0
Clarity: 1 (because of his commenting out it made sense during class) 

I used my class notes to help with this diagnostic.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,7 +5,19 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (originalArray, callback) {
+  const workingArray = originalArray.slice();
+  let minimum = array.reduce(min); //array isn't defined stated in grunt nag
+
+  if (typeof callback === 'undefined') {
+    minimum = workingArray.shift();
+  }
+
+  array.forEach(function(originalArray, callback) {
+    minimum = callback(minimum, originalArray, callback);
+  });
+
+  return minimum;
 
 };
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -48,9 +63,10 @@ const select = function (/* array, callback */) {
 // for which invoking the callback returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.find`
-const first = function (/* array, callback */) {
 
-};
+const first = function (workingArray, callback) {
+  let index = workingArray.findIndex(callback);
+
```
bernardlee: A missing closing brace to enclose the body of ```first()``` breaks the script here.

lib/diagnostic.js
```diff
@@ -5,7 +5,19 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (originalArray, callback) {
```
bernardlee: This function is being passed into ```reduce()``` as a callback function.  Its parameters should be the accumulator (or ```previous``` element in this case), the ```current``` element, and the current working array (omitted here).  In this particular case, ```previous``` and ```current``` are assumed to be numbers.

lib/diagnostic.js
```diff
@@ -5,7 +5,19 @@
...
+  const workingArray = originalArray.slice();
```
bernardlee: Since we're expecting both arguments to be numbers, calling ```slice()```, an array method, on a number will throw an error.

lib/diagnostic.js
```diff
@@ -14,7 +26,10 @@ const min = function (/* previous, current */) {
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
+const sum = function (originalArray, callback) {
```
bernardlee: This function has the same issues as ```min()```.

bernardlee:
0/5

PR #154, opened 2017-01-19
Response

dsweetser:


lib/diagnostic.js
```diff
@@ -5,17 +5,18 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
-
+const min = function (previous, current) {
```
bernardlee: This function needs a return value.

lib/diagnostic.js
```diff
@@ -29,8 +30,12 @@ const sum = function (/* previous, current */) {
 // in the array returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.some`
-const any = function (/* array, callback */) {
-
+const any = function (array, callback) {
+  for (let i = 0; i < array.length; i++) {
+    if (callback(array[i])) {
```
bernardlee: ```callback()``` should be called with three arguments, as the instructions have stated.

lib/diagnostic.js
```diff
@@ -39,8 +44,13 @@ const any = function (/* array, callback */) {
 // the callback returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.filter`
-const select = function (/* array, callback */) {
-
+const select = function (array, callback) {
+  let newArray = []
+  for (var i = 0; i < array.length; i++) {
+   if (any(array[i]) {
```
bernardlee: This line will break the script.  There needs to be an additional closing parenthesis.

bernardlee:
0/5

PR #156, opened 2017-01-19
Try Array Methods

skylarkJ:


lib/diagnostic.js
```diff
@@ -5,7 +5,20 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (/* previous, current */) {
+const min = function (previous, current) {
+  const workingArray = previous.slice();
```
bernardlee: ```previous``` and ```current``` are assumed to be numbers.  Since you're trying to invoke ```slice()```, an array method, on a number, this will throw an error.

lib/diagnostic.js
```diff
@@ -14,7 +27,20 @@ const min = function (/* previous, current */) {
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (/* previous, current */) {
+const sum = function (previous, current, callback) {
+  const workingArray = previous.slice();
```
bernardlee: This has the same issue as ```min()```.

lib/diagnostic.js
```diff
@@ -39,10 +65,12 @@ const any = function (/* array, callback */) {
 // the callback returns a truthy value
 // you may use any feature of JavaScript
 // other than `Array.prototype.filter`
-const select = function (/* array, callback */) {
-
+const select = function (array, callback) {
+    return !array.some((e, i, a)) => !callback(e, i, a));
```
bernardlee: ```some``` returns a boolean value.  We want this function to return an array.

bernardlee:
0/5

PR #158, opened 2017-03-26
Lm01/solution

gaand:


lib/diagnostic.js
```diff
@@ -15,7 +15,7 @@ const min = function (previous, current) {
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
 const sum = function (previous, current) {
-
+  return previousValue + currentValue
 }
```
payne-chris-r:
```diff
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-  return previousValue < currentValue ? previousValue : currentValue
+  return previous < current ? previous : current
 }

 // Question 2
@@ -15,7 +15,7 @@ const min = function (previous, current) {
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
 const sum = function (previous, current) {
-  return previousValue + currentValue
+  return previous + current
 }
```

payne-chris-r: Not sure what order these happened in, but my guess is you got it working and then rebased off of master? I'm not sure. What I DO know is that `previousValue` isn't defined unless we change the parameters to `previousValue` as well. On `master`, these are written as `...function (previous, current)`

payne-chris-r (commented):
Tests don't pass

lib/diagnostic.js
gaand: I'm not sure what went wrong.  I'm pushing a correction.

PR #160, opened 2017-03-29
Diagnostic for 3-29

awoodrum87:
0/5 ..obviously


benjimelito:
Looks like maybe you were able to get a solution for question 2. We're going to go over this stuff during workshop today, so no worries on not understanding a lot of it. 1/5

PR #161, opened 2017-03-29
Add diagnostic.js for js-array-iteration-diagnostic

jkjellan:
I wrote two callback functions to be passed to 'reduce',
and I used javascript methods 'every' and 'forEach' to
write my own functions for the last 3 quetions.  All
grunt tests pass

Still takes me awhile and sometimes I can't come up with the solution,
but the concepts are making sense

Comfort: 4
Clarity: 5

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,7 +6,7 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  return current < previous ? current : previous
 }
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -6,7 +6,7 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  return current < previous ? current : previous
```
benjimelito, line 9: Nice one-liner!

benjimelito:
Everything looks good, nice going. 5/5 🌈

PR #163, opened 2017-03-29
complete js array iteration methods diagnostic

buffalo0707:
Comfort: 3
Clarity: 2
I feel the way in which this unit was designed confused many of us.
While I understand the reasoning for wanting to show us how these array
methods work, I think it is more important at this point for us to know
how to use them. Additionally, Introducing them to us along with fat arrow 
formulasled to a lot of confusion over syntax, which distracted from the real
material being presented.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,10 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (previous, current) {
 
+const min = function (previous, current) {
+  if (current < previous) return current
+  return previous
 }
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -40,7 +44,8 @@ const hasNone = function (array, callback) {
 // you may use any feature of JavaScript
 // to solve this problem
 const last = function (array, callback) {
-
+  const newArray = array.filter(callback)
+  return newArray[newArray.length - 1]
```
benjimelito, line 48: Careful, what you are doing here is calling the function `newArray` inside your function called `newArray`. Also, you want to make sure that the callback function is actually returning `true`.

benjimelito:
I think a lot of your fellow developers feel similarly about this. We will be going over callbacks during workshop today, so feel free to ask any questions you might on this at that time. 4/5

PR #165, opened 2017-03-29
Added response to diagnostic

ghost:


lib/diagnostic.js
```diff
@@ -7,6 +7,20 @@
...
+
+    // create empty array
+    let minValue = []
+
+    // this assigns the previous array elements to the minValue
+    // array so that the original array is not modified
+    minValue[i] = previous[i]
+
+    // use the Math.min function to find the min value in an array.
+    current = Math.min(minValue)
```
benjimelito, line 21: I don't believe you can actually use `Math.min` on an array, you would have to pass it individual numbers as arguments.

lib/diagnostic.js
```diff
@@ -16,6 +30,16 @@ const min = function (previous, current) {
 // `let total = array.reduce(sum);`
 const sum = function (previous, current) {
 
+  // zero out current value
+  current = 0
+
+  // use an accumulator
+  for (let i = 0; i < previous.length; i++) {
```
benjimelito, line 37: This function will be passed to a `reduce` function, so there is no need to actually use a loop in here!

benjimelito:
No worries if you didn't quite understand this one, we will be going over it during workshop today. 0/5

PR #166, opened 2017-03-29
Submiting Diagnostic

sjruby:
Comfort 3/5
Confidince 3/5

It makes sense but I woudl need more time to finish this.


benjimelito:
No worries on not finishing, this seemed to be a difficult diagnostic for many people. Looks like you got the first two down, 2/5

PR #167, opened 2017-03-29
Response

andyjzhong:
Comfort: 1 / 5
Clarity: 1 / 5

I don't think my lack of comfort and clarity has to do with the
consultants not teaching it well, I think it's just the lack of time
spent on the subject. We went through those quite fast and I wish I had
more time in the evenings to review.

I think there is also some confusion caused by the fat arrow If
statements because I barely got to use and understand the If statement
syntax but we were already on to the =>. It's cool and quicker to write,
but would be nice if the callbacks were taught with the standard If
statement syntax and maybe later when we feel comfortable, teach it with
 =>'s.'

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,7 +6,16 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  // Set a dummy variable to store the result for later
+  const result = 0
+  // For each item in a provided array...
+  for (let i = 0, i < array.length; i++) {
+  // Loop through and compare index[0] and index[1]. If the result is greater than i,
+    if result > array[i] {
+    }
+  // Then store the new smaller number into the result.
+  return result
+  // Returns the minimum value in the array.
 }
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -6,7 +6,16 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  // Set a dummy variable to store the result for later
+  const result = 0
+  // For each item in a provided array...
+  for (let i = 0, i < array.length; i++) {
```
benjimelito, line 12: No need to do a loop inside here, since this function will be passed to `reduce` which does this for us.

lib/diagnostic.js
```diff
@@ -15,7 +24,17 @@ const min = function (previous, current) {
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
 const sum = function (previous, current) {
+  // Set a variable to store the result.
+  const result = 0
+  // Look at each item in the array
+  for (let i = 0, i < array.length; i++) {
```
benjimelito, line 30: Same as above ^^

benjimelito:
Hopefully yesterday's discussion cleared up some of your questions about arrow functions, if you have any other questions about array iterations please feel free to reach out to myself or any of the other consultants. 1/5

PR #168, opened 2017-03-29
Response: JS Array Iteration Methods

naributler:
Clarity: 3
Comfort: 3

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,8 +6,11 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -6,18 +6,26 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  return previous < current ? previous : current
```
benjimelito, line 9: Nice ternary operator!

benjimelito:
Nice work! For questions 3 & 5 we were looking for you to use the array iteration methods rather than a `for loop`, but looks like you were able to come up with some nice solutions either way. 4/5

PR #169, opened 2017-03-29
JS array iteration diagnostic

deeforte:
commit for js array iteration diagnostic
comfort: 1
clarity: 1

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,7 +6,13 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  let minimum = null
+  previous.reduce(function (minimum) {
+    if (previous[key] < minimum) {
+      minimum = previous[key], 0)
+    }
+  }
+  return result
 }
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -6,7 +6,13 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  let minimum = null
+  previous.reduce(function (minimum) {
```
benjimelito, line 10: Nice! Although you didn't have to write out this function as a `reduce`. The wording for this question might be a little unclear, but you were only required to provide the function that we would pass to a `reduce` function.

lib/diagnostic.js
```diff
@@ -15,7 +21,11 @@ const min = function (previous, current) {
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
 const sum = function (previous, current) {
```
benjimelito, line 23: Same as above, but nice work nonetheless

benjimelito:
Looks like you had good answers to the first two questions. Hopefully yesterday's review cleared some things up for you, but if not feel free to reach out to myself or Chris if you need more assistance. 2/5

PR #170, opened 2017-03-29
Diagnostic

brianedmunds:
My solutions

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -7,6 +7,13 @@
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
 
+  let result = arguments[0]
+  for (let i = 1; i < arguments.length; i++) {
+    if (arguments[i] < result) {
+      result = arguments[i]
+    }
+  }
+  return result
 }
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -7,6 +7,13 @@
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
 
+  let result = arguments[0]
```
benjimelito, line 10: We were looking for you to use an array iteration method rather than a `for loop` here, but this definitely works.

lib/diagnostic.js
```diff
@@ -16,7 +23,12 @@ const min = function (previous, current) {
 // `let total = array.reduce(sum);`
 const sum = function (previous, current) {
 
-}
+let total = 0
+let numbersArray = [5, 10, 15, 20, 25)]
```
benjimelito, line 27: Same as above ^^

benjimelito:
Looks like you had the right idea, but like I mentioned we were looking for you to use some of the array iteration methods rather than a `for loop`. Be sure to let us know if you need any extra time to go over this topic. 2/5

PR #171, opened 2017-03-29
Diagnostic Response

rrdaniels85:
Comfort: 2/5
Clarity: 3/5

I found this challenging. Just used class notes.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,7 +6,11 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  let minvalue = 0
+  for (let i = 1; i<Array.length; i++)
+  if (previous > current){
+    minValue = current
+  }
 }
 
 // Question 2
@@ -15,8 +19,12 @@ const min = function (previous, current) {
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
 const sum = function (previous, current) {
+  let newSum = previous
+  for (let i = 1; i<Array.length; i++){
+    newSum += current
+  }
+  }
 
-}
 
 // All of the callbacks passed to the functions
 // in questions 3 through 5 should be invoked with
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -6,7 +6,11 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  let minvalue = 0
+  for (let i = 1; i<Array.length; i++)
```
benjimelito, line 10: This would definitely work, but since this is a function that is actually going to be passed to reduce function there is no need for a loop.

lib/diagnostic.js
```diff
@@ -15,8 +19,12 @@ const min = function (previous, current) {
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
 const sum = function (previous, current) {
+  let newSum = previous
+  for (let i = 1; i<Array.length; i++){
```
benjimelito, line 23: Same as above ^

lib/diagnostic.js
```diff
@@ -40,7 +49,13 @@ const hasNone = function (array, callback) {
...
 // to solve this problem
 const last = function (array, callback) {
-
+  let newArray = []
+  for(let i = 0; i<array.length; i++)
+  if (callback(array[i])){
+    newArray.push(array[i])
+  }
+  lastOne = newArray.length
+  return newArray[lastOne-1]
```
benjimelito, line 58: Nice, this would definitely work

lib/diagnostic.js
```diff
@@ -51,9 +66,19 @@ const last = function (array, callback) {
 // you may use any feature of JavaScript
 // to solve this problem
 const hasOne = function (array, callback) {
-
+    let newArray = []
```
benjimelito, line 69: Cool 🕶

benjimelito:
Nice job finding some working solutions, however like I mentioned in the code we were looking for you use array iteration methods. 3/5

PR #172, opened 2017-03-29
Response - js-array-iteration-methods-diagnostic

rlarsonil71:
Comfort: 2
Clarity : 2

Need to practice callbacks - Use js-array-iteration-methods training

I was stumped on the problem instructions and have a feeling I made it to hard.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,16 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (previous, current) {
 
+const array = [1, 2, 4, 3]
+
+const min = function (previous, current) {
+  let minimum = reduce(array, (previous, current) => {
+    if (previous < current) {
+      return current
+    }
+  })
+  return minimum
 }
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,8 +5,16 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (previous, current) {
 
+const array = [1, 2, 4, 3]
+
+const min = function (previous, current) {
+  let minimum = reduce(array, (previous, current) => {
```
benjimelito, line 12: Cool 👍 . Although you don't have to have a `reduce` inside your function, since the function that you write will be passed to another `reduce` function per the prompt.

lib/diagnostic.js
```diff
@@ -15,7 +23,9 @@ const min = function (previous, current) {
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
 const sum = function (previous, current) {
-
+  let minimum = reduce(array, (previous, current) => {
+  })
```
benjimelito, line 27: Same as above ^^

lib/diagnostic.js
```diff
@@ -31,7 +41,10 @@ const sum = function (previous, current) {
 // you may use any feature of JavaScript
 // to solve this problem
 const hasNone = function (array, callback) {
+  let result = false
 
+  if (every(array, (element, index, a)) => result = true
```
benjimelito, line 46: Nice

benjimelito:
Looks like you had the right idea, but hopefully yesterday cleared up some of your questions about callbacks and array iteration methods. Let me know if you need anything else cleared up. 2/5

PR #173, opened 2017-03-29
Add responses to diagnostic

jim-moody:
Comfort: 4
Clarity: 4

The wording on question #5 was a little confusing, it's saying two different things.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,7 +6,7 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  return current < previous && current
 }
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -40,7 +40,7 @@ const hasNone = function (array, callback) {
 // you may use any feature of JavaScript
 // to solve this problem
 const last = function (array, callback) {
-
+  return array.reverse().find((e, i, a) => callback(e, i, a))
```
benjimelito, line 43: Cool way of solving it

benjimelito: Although I don't know if this checks to make sure the return value is `true`

benjimelito:
Looks good for the most part, nice going. In #5 we were just asking to return the last element in an array that returns true if the callback is executed on it. 3/5

PR #174, opened 2017-03-29
Fallon response

jonifallon:
Still don't understand syntax.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,7 +6,8 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+      return Math.min.apply( Math, array );
+  };
 }
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -31,7 +32,10 @@ const sum = function (previous, current) {
 // you may use any feature of JavaScript
 // to solve this problem
 const hasNone = function (array, callback) {
-
+  for (let i = 0; i < array.length; i++) {
+    if (callback(array[i], i, array)) return i
```
benjimelito, line 36: You want this to return either `true` or `false`, not an index

lib/diagnostic.js
```diff
@@ -40,6 +44,10 @@ const hasNone = function (array, callback) {
 // you may use any feature of JavaScript
 // to solve this problem
 const last = function (array, callback) {
+  for (let i = 0; i < array.length; i++) {
+    if (callback(array[i], i, array)) return i
```
benjimelito, line 48: Same as above ^^

benjimelito:
Let me or another consultant know if you're still having trouble with the array iteration method syntax, I hope some things were cleared up for you yesterday. 2/5

PR #175, opened 2017-03-29
Submitted Diagnostic

jscohen:


lib/diagnostic.js
```diff
@@ -5,17 +5,27 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (previous, current) {
-
+const min = function (previous, originalArray) {
```
benjimelito, line 8: Nice, this is how you wold do it with a `for loop`. However, we're looking for a function that will be passed to a reduce function

lib/diagnostic.js
```diff
@@ -5,17 +5,27 @@
...
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (previous, current) {
-
+const sum = function (previous, originalArray) {
+  let sum = originalArray[0]
```
benjimelito, line 24: Same as above ^^

benjimelito:
Looks like you came up with some good solutions to a few of the answers, but not necessarily in the format we were looking for. Good work nonetheless - 3/5

PR #176, opened 2017-03-29
Add response

khan-ngo:


khan-ngo:
2/2 

Had to refer back to notes, but I'm not clear.

benjimelito:
You were definitely on the correct path in the first two questions. Let me know if you still need some clarification on array iteration methods. 1/5

PR #177, opened 2017-03-29
js-array-iterations diagnostic so hard

MattKwonDo:
this was really hard, definitely needed more time to work on this, will
be practicing tonight

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,17 +5,33 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (previous, current) {
-
+const min = function (array, callback) {
+  const minimum = Math.min[array] // array.reduce(callback)
+  return minimum
 }
+//min([-1, -2, -3, -4, -5], callback(e, i, a))
+
+let min = function (array) {
+  let minimum = array[0]
+  for (let i = 1; i < array.length; i++) {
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,33 @@
...
-const min = function (previous, current) {
+const min = function (array, callback) {
+  const minimum = Math.min[array] // array.reduce(callback)
+  return minimum
+}
+//min([-1, -2, -3, -4, -5], callback(e, i, a))
 
+let min = function (array) {
+  let minimum = array[0]
+  for (let i = 1; i < array.length; i++) {
```
benjimelito, line 16: This would work, but no need to use a `for loop` inside here since we're going to be passing this function to a `reduce` function.

lib/diagnostic.js
```diff
@@ -5,17 +5,33 @@
...
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (previous, current) {
-
+const sum = function (array, callback) {
+  const total = array.reduce(callback(e, i, a))
+  return total
```
benjimelito, line 34: Make sure you increment this

benjimelito:
Seems like you had some of the concepts correct. Hopefully yesterday's review clarified a few things for you, but if you still need some rehashing please let me know. 1/5

PR #178, opened 2017-03-29
completed only 3 of the five questions

gastevend:
with the third being REAL shaky


benjimelito:
Really good looking stuff for the first 3, looks like you simply may have needed more time. 3/5

PR #179, opened 2017-03-29
Array iteration methods

antleo1995:
Handled the first two easily. The 3rd one, as expected really tripped me
up. This is not a concept I am totally comfortable with.

Gonna probably call it a 2 and a 2.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,7 +6,11 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  let result = previous
+  if (previous > current) {
+    result = current
+  }
+  return result
 }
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -31,7 +35,10 @@ const sum = function (previous, current) {
 // you may use any feature of JavaScript
 // to solve this problem
 const hasNone = function (array, callback) {
-
+  array.forEach((e, i, a) => {
+    if (array.indexOf(callback(e, i, a)) === -1) { return true }
+    else { return false }
```
benjimelito, line 40: Close, but since this is inside a `forEach` loop it would actually return either `true` or `false` for every element in the array

benjimelito:
Really nice job on the first two, close on the third. No stress, this diagnostic was very difficult for most people in the cohort. 2/5

PR #182, opened 2017-03-29
JS Array Iteration Method Diagnostic

cpm5041:
Comfort: 1
Clarity: 1

I struggled with this one. I understand the basic concepts but have
trouble implementing. I will need to spend more time on this and
will probably need guidance during workshop time.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,7 +6,16 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  // assign a minimum value variable to return
+  const minValue = previous
+  // for every value in an array of an arbitrary length
+  for (let i = 0; arguments.length; i++) {
+    if (current < previous) {
+// set the current value to the min value if it is less than previous
+  current = minValue
+// return the min value
+  } return minValue
+  }
 }
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -6,7 +6,16 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  // assign a minimum value variable to return
+  const minValue = previous
+  // for every value in an array of an arbitrary length
+  for (let i = 0; arguments.length; i++) {
```
benjimelito, line 12: No need to loop here, since we're inside a function that is going to be passed to a `reduce` function

lib/diagnostic.js
```diff
@@ -15,7 +24,16 @@ const min = function (previous, current) {
...
 // `let total = array.reduce(sum);`
 const sum = function (previous, current) {
-
+// start with the sum at the zero
+  let sumOfArray = 0
+  let previous = initialValue
+  // for every value in an array of an arbitrary length
+  for (let i = 0; arguments.length; i++) {
+    // add the values of the arguments to the sum
+    sumOfArray = initialValue + current
```
benjimelito, line 33: This would definitely work, but see above ^^

benjimelito:
Hope that the review yesterday cleared some things up for you, if not then definitely let someone know. This was a hard topic for a lot of people, so don't stress too much about it. 2/5

PR #184, opened 2017-03-29
Add response

lisamsmith100:


lib/diagnostic.js
```diff
@@ -1,21 +1,45 @@
...
 const min = function (previous, current) {
 
+  //for the callbacks, these are called for each element, correct?
+
+  let result = previous
+
+  if (result < current) {
+    return result
+  } else {
+    result = current
```
benjimelito, line 26: Cool 😎

lib/diagnostic.js
```diff
@@ -1,21 +1,45 @@
...
 }
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
+
+// i don't understand for sure if previous or current are arrays or values in
+// array; just discovering my questions
```
benjimelito, line 38: So `previous` is sort of the accumulation of all items up until this point, and `current` is simply the current value.

benjimelito:
Looks like you had a pretty good start on this one. Let me know if you still feel shaky on any concepts here. Also, please make sure to get these in before 9:30am!

PR #186, opened 2017-03-29
JavaScript Array Methods Diagnostic

octater:
NFC

Comfort: 0
Clarity: 0

I used my class notes to help with this diagnostic.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,7 +6,13 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  previous.forEach(function callback(currentValue, index, array) {
+    current = previous[0]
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -6,7 +6,13 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  previous.forEach(function callback(currentValue, index, array) {
+    current = previous[0]
+    for (let i = 1; i < previous.length; i++) {
```
benjimelito, line 11: This would definitely work on its own, but we don't need to use a `for loop` since this function is going to be passed to a `reduce` function.

benjimelito:
Hopefully yesterday cleared up a few things for you, but please reach out if you are still unclear on anything. A lot of people struggled on this one, so no stress. Please make sure to get these in before 9:30am though!

PR #187, opened 2017-03-29
Done in more ways then one

Ryaong792:
fist of 5....1 or 0.  I was still trying to catch up with understanding
reduce fat arrows etc.  Was stuck on the first question and started
to move on to the 2nd then time was done.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,7 +6,9 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  let ..
+  let minimum = minimun.reduce(min)
+  return min
 }
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -15,8 +17,11 @@ const min = function (previous, current) {
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
 const sum = function (previous, current) {
+  let total = sum.reduce(function (prevous, current) {
```
benjimelito, line 20: Yup. You didn't need to write out the entire function for this one however, simply the callback function that is passed to the `reduce`.

benjimelito:
A lot of people struggled with this topic, so try not to stress too much about it. Feel free to reach out if you're still unclear on anything. Also, please try to get these in before 9:30am even if you're not finished.

PR #188, opened 2017-03-29
Angela M.'s response

angmas:
comfort: 3
clarity: 3

I just need more time to think through and work through the diagnostic questions. The diagnostics questions are a great way for me to exercise what I know and to try out different things. It's just not enough time to figure it all out. The one thing that absolutely slowed down my learning was the introduction of the fat arrow syntax ON TOP OF some complex lessons on abstraction. THis is like getting people who are used to driving automatic into a manual trans car to teach them how to learn how to drive to a place. The lesson is how to get somewhere specific, but people could be stuck on learning the stick transmission.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,7 +6,9 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  if (current < previous) {
+    return current
+  } else { return previous }
 }
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -6,7 +6,9 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  if (current < previous) {
+    return current
+  } else { return previous }
```
benjimelito, line 11: 👍

benjimelito:
You came up with some good-looking answers to the questions that you were able to get to, but please try to get these in before 9:30am even if you aren't finished! Also, hopefully you are clearer on some of these concepts after yesterday, but please let me or someone else know if you need extra assistance

PR #189, opened 2017-04-03
Completed all questions after more review.

danielleletarte:
All grunt tests passed.


benjimelito:
Awesome, glad you nailed these after some review 👍

PR #190, opened 2017-04-06
JS iterations response

Maggicorp:
commit first two answers

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,12 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (previous, current) {
 
+const min = function (previous, current) {
+  if (previous > current) {
+    previous = current
+    return previous
+  }
 }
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -31,7 +36,12 @@ const sum = function (previous, current) {
 // you may use any feature of JavaScript
 // to solve this problem
 const hasNone = function (array, callback) {
-
+  const newArray = array.filter(callback)
```
benjimelito, line 39: Nice solution 👍

lib/diagnostic.js
```diff
@@ -40,7 +50,9 @@ const hasNone = function (array, callback) {
 // you may use any feature of JavaScript
 // to solve this problem
 const last = function (array, callback) {
-
+  const newArray = array.forEach(callback)
```
benjimelito, line 53: Right idea, but remember that `forEach` does not automatically create a new array for you. `Map` would probably be a better option.

benjimelito:
Looks good! Nice work, 4/5

PR #194, opened 2017-04-06
Anwering questions

Anddrew914:


benjimelito:
Nice job, I'm sure you would've gotten 💯  if you had more time. 4/5

PR #197, opened 2017-04-06
Push Request for js-array-iteration methods

noobiwankenoobi:
I put in some placeholder code on a couple of the questions
that I would then work with to try and solve them.
I am not fast enough with this particular material yet to do much
more than that in a half hour. I still need to practice starting
from scratch on this material. We are often shown where to start.
Having a variety of questions and trying to figure out which methods to use on my own has proven challenging.
I will do extra practice today.

Comfort: 2
Clarity:3

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -26,12 +26,18 @@ const sum = function (previous, current) {
 
 // Question 3
 // Write a function that returns `true`
-// if an only if invoking the callback
+// if and only if invoking the callback
 // for every element in the array returns `false`
 // you may use any feature of JavaScript
 // to solve this problem
 const hasNone = function (array, callback) {
-
+    array.forEach((e, i, a) => {
+      if ((callback(e, i, a)) === false) {
+        return true
+      }
+    })
+    return result
+  }
 }
 
 // Question 4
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -26,12 +26,18 @@ const sum = function (previous, current) {
...
 // you may use any feature of JavaScript
 // to solve this problem
 const hasNone = function (array, callback) {
-
+    array.forEach((e, i, a) => {
+      if ((callback(e, i, a)) === false) {
+        return true
+      }
+    })
+    return result
```
benjimelito, line 39: Make sure you're defining this variable somewhere before returning it!

lib/diagnostic.js
```diff
@@ -40,12 +46,18 @@ const hasNone = function (array, callback) {
 // you may use any feature of JavaScript
 // to solve this problem
 const last = function (array, callback) {
+    let newArray = []
+    newArray = array.filter(callback){
 
+      }
+    })
+    return newArray[(newArray.length - 1)]
+  }
```
benjimelito, line 55: 👍

benjimelito:
Looks like you were able to get a solution to one of the harder questions. I would've probably recommend starting with some of the easier questions at the top before tackling the more difficult ones, but like you said you'll also probably become more comfortable at this sort of thing with practice. 1/5

PR #198, opened 2017-04-06
Response

mlynnes:
I ran out of time trying to solve these problems.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -13,10 +13,10 @@ const min = function (previous, current) {
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
-// `let total = array.reduce(sum);`
-const sum = function (previous, current) {
-
-}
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -13,10 +13,10 @@ const min = function (previous, current) {
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
-// `let total = array.reduce(sum);`
-const sum = function (previous, current) {
-
-}
+// let total = array.reduce(sum)
+const sum = .reduce(function (previous, current) {
```
benjimelito, line 17: Yup, but you actually didn't need to make this a reduce function. They were simply asking for the function that you would pass to reduce.

benjimelito:
Looks like you got a good solution for one question at least. I would imagine that you'll get faster with this sort of thing as you gain experience, but please let me know if you need any extra review on these concepts. 1/5

PR #199, opened 2017-04-06
Complete JS Diagnostic

sstone72389:
Comfort: 1
Clarity: 2

I used Google and my class notes. This was very tough

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,7 +6,7 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+Ran out of time
 }
 
 // Question 2
@@ -14,8 +14,11 @@ const min = function (previous, current) {
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (previous, current) {
 
+const sum = function (array) {
+  array.reduce(function(prev, cur) {
+    return prev + cur;
+  }, 0);
 }
 
 // All of the callbacks passed to the functions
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -6,16 +6,19 @@
...
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (previous, current) {
 
+const sum = function (array) {
+  array.reduce(function(prev, cur) {
+    return prev + cur;
+  }, 0);
```
benjimelito, line 21: Nice 👍

lib/diagnostic.js
```diff
@@ -30,16 +33,24 @@ const sum = function (previous, current) {
 // for every element in the array returns `false`
 // you may use any feature of JavaScript
 // to solve this problem
-const hasNone = function (array, callback) {
 
+const hasNone = function (array, callback) {
+    const isFalse = array.filter(hasNone = function(array) {
+    if ((callback(element, i, originalArray) === false)) {
+  }
```
benjimelito, line 40: You want to make sure that a function you are passing to `filter` either returns `true` or `false`

lib/diagnostic.js
```diff
@@ -50,8 +61,13 @@ const last = function (array, callback) {
...
 // you may use any feature of JavaScript
 // to solve this problem
-const hasOne = function (array, callback) {
 
+const hasNone = function (array, callback) {
+    const isFalse = array.filter(hasNone = function(array) {
+    if ((callback(element, i, originalArray) === true)) {
+  }
+    return true
+  }
```
benjimelito, line 70: Same as the above function ^^

benjimelito:
Let me know if you have any lingering questions on this topic, I think it will get more comfortable with practice. 2/5

PR #200, opened 2017-04-06
JS Array Iteration Diagnostic

bcem2004:
comfort: 0-1
clarity: 0-1

I am able to follow along when these are being explained, but I am not at a level where I can produce them on my own yet. I didn't understand what questions 3-5 were asking. I will try to get some 1-1 time with a consultant. All answers provided were a combination of examples Slacked from Antony and Ben and Stack Overflow via Google.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,16 +5,24 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (previous, current) {
 
+function arrayMin(arr) {
+  const len = arr.length, min = Infinity;
+  while (len--) {
+    if (arr[len] < min) {
+      min = arr[len];
+    }
+  }
+  return min;
 }
-
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,16 +5,24 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (previous, current) {
 
+function arrayMin(arr) {
+  const len = arr.length, min = Infinity;
```
benjimelito, line 10: Interesting solution, I guess comparing something to `Infinity` at the start works

lib/diagnostic.js
```diff
@@ -5,16 +5,24 @@
...
-
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (previous, current) {
+const sum = [0, 1, 2, 3].reduce(function (a, b) {
+  return a + b;
+}, 0);
```
benjimelito, line 25: Cool 👍

benjimelito:
This is the sort of thing that you will most likely become more comfortable with as you practice. That being said, let me (or another consultant) know if you would like to go over certain topics more in depth. 2/5

PR #201, opened 2017-04-06
iteration

eliottenos:
Comfort: 2
Clarity: 2

Used stackoverflow, mozilla, notes and class material.

Ran out of time overthinking and looking functions up.  Much more confident
than a day ago but need practice obviously!


benjimelito:
Yup, you'll definitely get quicker with more practice. Glad to hear you're gaining confidence, please reach out if you need any extra assistance! 1/5

PR #202, opened 2017-04-06
First test passing, spelled branch name wrong, fixed.

aprilsourz:


aprilsourz:
comfort: 2
clarity: 2

I looked on stack overflow.

lib/diagnostic.js
```diff
@@ -6,7 +6,7 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
+  return Math.min(previous, current)
```
benjimelito, line 9: Nice 👍

lib/diagnostic.js
```diff
@@ -15,7 +15,7 @@ const min = function (previous, current) {
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
 const sum = function (previous, current) {
-
+  return Math.max(previous, current)
```
benjimelito, line 18: This will return the largest element in an array, but you actually want to be adding them together. Something like `return previous + current`

benjimelito:
Looks like you are grasping the idea of array iteration methods, but let me know if you need to go over some of these concepts further. 1/5

PR #203, opened 2017-04-06
Added my responses to lib/diagnostic.js

ghost:
comfort: 0
clarity: 0

I looked at class notes but could not translate them to the current problems, in the time alloted, at my skill level these would take me a whole night to diagnose the questions asked.


benjimelito:
Let me (or another consultant) know if you'd like to go over some of these concepts more in depth, they're definitely not easy to master. 0/5

PR #206, opened 2017-04-06
array iteration methods diagnostic

xiangcatherine:
Completed questions 1-4.

Comfort: 3
Clarity: 3

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -49,7 +49,11 @@ const hasNone = function (array, callback) {
 // you may use any feature of JavaScript
 // to solve this problem
 const last = function (array, callback) {
-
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -31,16 +34,26 @@ const sum = function (previous, current) {
 // you may use any feature of JavaScript
 // to solve this problem
 const hasNone = function (array, callback) {
-
-}
+  array.forEach((e, i, a) => {
```
benjimelito, line 37: So this would actually return `true` or `false` for each element. What you want is a function that will return EITHER `true` or `false`

benjimelito:
Nice job on the first two questions, if you need some more help with certain topics let me know. 2/5

PR #207, opened 2017-04-06
Response

mangoceylon:
Clarity: 4
Comfort: 3

Used google and notes from the class.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,8 +5,8 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (previous, current) {
-
+const min = function (a, cV, cI, array) {
+  return Math.min(a, cV)
 }
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,17 +5,17 @@
 // the minimum value in an array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
-const min = function (previous, current) {
-
+const min = function (a, cV, cI, array) {
```
benjimelito, line 8: This will totally work, but functions passed to a `reduce` should only have 2 arguments

lib/diagnostic.js
```diff
@@ -5,17 +5,17 @@
...
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (previous, current) {
-
+const sum = function (a, cV) {
+  return a + cV
```
benjimelito, line 18: Yup

benjimelito:
Nice job on the first two, let me know if you need more review on some of these topics. 2/5

PR #208, opened 2017-04-06
  response

JLisaW:


lib/diagnostic.js
```diff
@@ -5,19 +5,21 @@
...
 
 // Question 2
 // Write a callback for reduce that produces
 // the sum of the values in the array
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let total = array.reduce(sum);`
-const sum = function (previous, current) {
-
+const sum = array.reduce(previous, current) {
+  current[previous] = true
```
benjimelito, line 18: Not sure what this line does? In any case I don't think you need it

lib/diagnostic.js
```diff
@@ -50,8 +53,8 @@ const last = function (array, callback) {
 // for one and only one element
 // you may use any feature of JavaScript
 // to solve this problem
-const hasOne = function (array, callback) {
-
+const hasOne = arr.filter((array, callback)) {
+  return true
```
benjimelito, line 57: You want to make sure that the `filter`ed array is the same size as the original here

benjimelito:
Solid work on the first two answers, 2/5

PR #211, opened 2017-04-06
Still confused on why a callback would be used in

rowlandhill:
I'm still unclear on why I would use a callback to find the sum.  I was confused on why I wouldn't just find the sum.  I still need quite a bit of review on this :)

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -6,17 +6,24 @@
 // this callback will be passed by the tests to the reduce method, e.g.
 // `let minimum = array.reduce(min);`
 const min = function (previous, current) {
-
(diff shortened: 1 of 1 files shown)


rowlandhill:
Update: after talking to peers and reviewing the solution branch, I see that I completely misunderstood what the diagnostic was asking me to do.  I thought I was supposed to be creating the entire function from scratch, and not just the callback method that would be used as an parameter/argument in an iteration method.  I definitely still need to review, but I misread/misunderstood the directions :/

benjimelito:
No worries about misunderstanding the directions, I think a lot of people did actually. If you need any more review please let me (or another consultant) know! 1/5

